PR #2, opened 2019-10-20
feat: crypto interface

jacobheun:
This adds a crypto interface to js libp2p, we currently don't have one. This also includes a mock crypto file for tests, which demonstrates how a crypto module might be built, including a handshake and stream wrapping which can be used for crypto boxing/unboxing.

Aside from the clear change to async iterables, this also updates the "existing" interface to use the `protocol` property instead of `tag` to be clearer about its purpose.

I added the iterable wrapper logic to the js-libp2p docs in a PR, in case it's useful to others https://github.com/libp2p/js-libp2p/pull/466


src/crypto/README.md
```diff
@@ -0,0 +1,68 @@
+interface-crypto
+==================
+
+> A test suite you can use to implement a libp2p crypto module. A libp2p crypto module is used to ensure all exchanged data between two peers is encrypted.
+
+**Modules that implement the interface**
```
mkg20001: Is plaintext not going to be a module?

src/crypto/README.md
```diff
@@ -0,0 +1,68 @@
...
+## API
+
+- `Crypto`
+  - `protocol<string>`: The protocol id of the crypto module.
+  - `secureInbound<function(PeerId, duplex)>`: Secures inbound connections.
+  - `secureOutbound<function(PeerId, duplex, PeerId)>`: Secures outbound connections.
+
+### Secure Inbound
+
+- `const { conn, remotePeer } = await crypto.secureInbound(localPeer, duplex)`
```
mkg20001: Earlier there was the possibility of streaming data to the connection _before_ the crypto was setup
Should we honor that and simply return an additional `upgraded` variable that contains a promise that resolves after the upgrade? (The error in that case would also bubble-up via read-calls)
Or should we block the flow until crypto is finished?

src/crypto/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Secures an inbound [streaming iterable duplex][iterable-duplex] connection. It returns an encrypted [streaming iterable duplex][iterable-duplex], as well as the [PeerId][peer-id] of the remote peer.
+
+**Parameters**
+- `localPeer` is the [PeerId][peer-id] of the receiving peer.
```
mkg20001: This should be optional, as we don't know it for incoming dials sometimes.

src/crypto/README.md
```diff
@@ -0,0 +1,68 @@
...
+  - `secureInbound<function(PeerId, duplex)>`: Secures inbound connections.
+  - `secureOutbound<function(PeerId, duplex, PeerId)>`: Secures outbound connections.
+
+### Secure Inbound
+
+- `const { conn, remotePeer } = await crypto.secureInbound(localPeer, duplex)`
+
+Secures an inbound [streaming iterable duplex][iterable-duplex] connection. It returns an encrypted [streaming iterable duplex][iterable-duplex], as well as the [PeerId][peer-id] of the remote peer.
+
+**Parameters**
```
mkg20001: The inbound doesn't seem to have a `remotePeer` argument

It might be useful to include that as an optional parameter. It could be useful in case a connection gets upgraded from one transport to another, to verify both sides didn't encounter a MITM.

src/crypto/README.md
```diff
@@ -0,0 +1,68 @@
...
+- `localPeer` is the [PeerId][peer-id] of the receiving peer.
+- `duplex` is the [streaming iterable duplex][iterable-duplex] that will be encryption.
+- `remotePeer` is the [PeerId][peer-id] of the remote peer. If provided, implementations **should** use this to validate the integrity of the remote peer.
+
+**Return Value**
+- `<object>`
+  - `conn<duplex>`: An encrypted [streaming iterable duplex][iterable-duplex].
+  - `remotePeer<PeerId>`: The [PeerId][peer-id] of the remote peer.
```
mkg20001: Possible a note should be added that this should usually be the same id as the remotePeer provided via arguments, just to make that a bit clearer?
(Could also be too much, I'll leave that up to you)

src/crypto/README.md
jacobheun: Yes, that's the intent, I'm planning on writing that and the private network refactor this week. I was planning on just bundling plaintext 2 with libp2p though, and linking to it here.

src/crypto/README.md
jacobheun: It's fully blocked now with async. A connection is not returned until crypto and muxer upgrading has been completed.

src/crypto/README.md
jacobheun: That's a good point, I will update to include that as optional.

jacobheun:
I've updated the docs and added a test for the expected `remotePeer` on an inbound connection. I also added an `errors` export with an `UnexpectedPeerError` as an initial value. We can add to this as we determine other, common errors crypto modules should throw.

src/crypto/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Secure Outbound
+
+- `const { conn, remotePeer } = await crypto.secureOutbound(localPeer, duplex, remotePeer)`
+
+Secures an outbound [streaming iterable duplex][iterable-duplex] connection. It returns an encrypted [streaming iterable duplex][iterable-duplex], as well as the [PeerId][peer-id] of the remote peer.
+
+**Parameters**
+- `localPeer` is the [PeerId][peer-id] of the receiving peer.
+- `duplex` is the [streaming iterable duplex][iterable-duplex] that will be encryption.
```
vasco-santos:
```suggestion
- `duplex` is the [streaming iterable duplex][iterable-duplex] that will be encrypted.
```

src/crypto/README.md
mkg20001: Ok, perfect. I'll keep this discussion open until that's resolved for tracking.

mkg20001:
I assume that with `js-interfaces` now being a thing, `interface-connection` is no longer needed and the functions return "just streams"
Is that correct?

jacobheun:
> I assume that with js-interfaces now being a thing, interface-connection is no longer needed and the functions return "just streams"
Is that correct?

Yes, that's correct. Now when you dial in libp2p the `Connection` object is not created until just before the dial returns. It's now basically just a metadata/convenience wrapper around the muxer and transport connection (`MultiaddrConnection`).

src/crypto/README.md
jacobheun: I also created https://github.com/libp2p/js-interfaces/issues/3 to track adding plaintext.

PR #7, opened 2019-11-07
feat: topology

vasco-santos:
Added topology interface


src/topology/index.js
```diff
@@ -0,0 +1,108 @@
+'use strict'
+
+const assert = require('assert')
+
+class Topology {
```
jacobheun: This really isn't a base Topology, it's a MulticodecTopology which should extend a base. Not all topologies are going to care about multicodecs or protocol changes. Two immediate use cases that come to mind for Topologies are Priority Peers and Bootstrap Nodes, which I would categorize as a PeerSet Topology. 

Let's say a js-ipfs browser node would like to stay connected to a Preload Node, `QmPreload`. Currently, libp2p doesn't handle that. What could happen, is a new PeerSet Topology could be created, which we should be able to achieve with the Base Topology.

```js
const priorityTopology = new Topology()
priorityTopology.peers.set('QmPreload', preloadPeerInfo) // We could iterate over several peers.
priorityTopology.min = priorityTopology.peers.size // We want to be connected to all our peers

// ... registration
```

This same setup could be used for Bootstrap nodes, instead of it being a "discovery" service. The only difference is that the min can be 0, because we shouldn't really care if we're connected to them, as long as our node has at least `ConnectionManager.min` connections, which will be handled elsewhere.

vasco-santos: Added `multicodec-topology`

src/topology/index.js
```diff
@@ -0,0 +1,47 @@
...
+    this.peers = new Map()
+  }
+
+  /**
+   * Notify about peer disconnected event.
+   * @param {PeerInfo} peerInfo
+   * @param {Error} [error]
+   * @returns {void}
+   */
+  disconnect (peerInfo, error) {
```
jacobheun: What's your thought behind passing error here? I don't think we're propagating disconnect errors to the Registrar and I'm not sure they're needed.

src/topology/multicodec-topology.js
```diff
@@ -0,0 +1,89 @@
...
+    max,
+    multicodecs,
+    handlers
+  }) {
+    super({ min, max, handlers })
+
+    assert(multicodecs, 'one or more multicodec should be provided')
+
+    this.multicodecs = Array.isArray(multicodecs) ? multicodecs : [multicodecs]
+    this.peers = new Map()
```
jacobheun: This is done in the base class.
```suggestion
```

src/topology/index.js
```diff
@@ -0,0 +1,47 @@
...
+   * @param {function} props.handlers.onConnect protocol "onConnect" handler
+   * @param {function} props.handlers.onDisconnect protocol "onDisconnect" handler
+   * @constructor
+   */
+  constructor ({
+    min = 0,
+    max = Infinity,
+    handlers
+  }) {
+    assert(handlers, 'the handlers should be provided')
```
jacobheun: I'm thinking we shouldn't require the handlers. Some topologies might not care, like the PeerSet Topologies. Multicodec needs them though. Maybe they should be required there and default to noop functions here?

src/topology/tests/multicodec-topology.js
```diff
@@ -0,0 +1,96 @@
...
+      const id2 = await PeerId.createFromJSON(peers[1])
+      const peer2 = await PeerInfo.create(id2)
+
+      const peerStore = topology._registrar.peerStore
+      peerStore.emit('change:protocols', {
+        peerInfo: peer2,
+        protocols: Array.from(topology.multicodecs)
+      })
+
+      await delay(500)
```
jacobheun: Are the delays in this suite actually needed? I don't think the event emitter is async. If I am mistaken then `delay(1)` should be sufficient.

src/topology/index.js
vasco-santos: That's for identifying if the `disconnect` was due to an error or not. I guess we can remove it until it can be useful

src/topology/index.js
vasco-santos: I think you are right, I will do it

src/topology/index.js
jacobheun: I think if we remove this and fill in the readme this should be gtg!

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
+interface-topology
+========================
+
+> Implementation of the topology interface used by the `js-libp2p` registrar.
+
+This interface has two main purposes. It uniforms the registration of libp2p protocols and enables a smarter connection management.
```
jacobheun: I think something like this would be good for now, and we can point to a more in depth overview once that's ready.

```suggestion
Topologies can be used in conjunction with `js-libp2p` to help shape its network and the overlays of its subsystems, such as pubsub and the DHT.
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+  max: 50,
+  multicodecs: ['/echo/1.0.0'],
+  handlers: {
+    onConnect: (peerInfo, conn) => {},
+    onDisconnect: (peerInfo) => {}
+  }
+})
+
+// Needs to set registrar in order to listen for peer changes
+topology.registrar = registrar
```
jacobheun:
```suggestion
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+  min: 0,
+  max: 50,
+  multicodecs: ['/echo/1.0.0'],
+  handlers: {
+    onConnect: (peerInfo, conn) => {},
+    onDisconnect: (peerInfo) => {}
+  }
+})
+
+// Needs to set registrar in order to listen for peer changes
```
jacobheun:
```suggestion
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+const toplogy = new MulticodecTopology({
+  min: 0,
+  max: 50,
+  multicodecs: ['/echo/1.0.0'],
+  handlers: {
+    onConnect: (peerInfo, conn) => {},
+    onDisconnect: (peerInfo) => {}
+  }
+})
+
```
jacobheun: I would just delete these 3 lines. We can explain the registrar property in the API section.
```suggestion
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## API
+
+The `MulticodecTopology` extends the `Topology`, which makes the `Topology` API a subset of the `MulticodecTopology` API.
+
+###  Topology
+
+- `Topology`
+  - `peers.set<function(id, PeerInfo)>`: Sets a peer in the topology.
```
jacobheun: Express this as a Map.
```suggestion
  - `peers<Map<string, PeerInfo>>`: A Map of peers belonging to the topology.
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+  - `disconnect<function(PeerInfo)>`: Disconnects a peer from the topology.
```
jacobheun:
```suggestion
  - `disconnect<function(PeerInfo)>`: Called when a peer has been disconnected
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+- `topology.disconnect(peerInfo)`
+
+**Parameters**
+- `peerInfo` is the [PeerInfo][peer-info] of the peer disconnected.
+
+###  Multicodec Topology
+
+- `MulticodecTopology`
+  - `registrar<Registrar>`: Sets the registrar in the topology.
```
jacobheun:
```suggestion
  - `registrar<Registrar>`: The `Registrar` of the topology. This is set by the `Registrar` during registration.
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+  - `peers.set<function(id, PeerInfo)>`: Sets a peer in the topology.
```
jacobheun: Let's just show that this is a Map and can be used accordingly. 
```suggestion
  - `peers<Map<string, PeerInfo>>`: The Map of peers that belong to the topology
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+    - `onDisconnect` is a `function` called everytime a peer is disconnected in the topology context.
+
+#### Set a peer
+
+- `topology.peers.set(id, peerInfo)`
+
+Add a peer to the topology.
+
+**Parameters**
+- `id` is the `b58string` that identifies the peer to add.
```
jacobheun: It will eventually be base32, so maybe just leave it generic
```suggestion
- `id` is the `string` that identifies the peer to add.
```

src/topology/README.md
```diff
@@ -0,0 +1,150 @@
...
+**Parameters**
+- `properties` is an `Object` containing the properties of the topology.
+  - `min` is a `number` with the minimum needed connections (default: 0)
+  - `max` is a `number` with the maximum needed connections (default: Infinity)
+  - `multicodecs` is a `Array<String>` with the multicodecs associated with the topology.
+  - `handlers` is an optional `Object` containing the handler called when a peer is connected or disconnected.
+    - `onConnect` is a `function` called everytime a peer is connected in the topology context.
+    - `onDisconnect` is a `function` called everytime a peer is disconnected in the topology context.
+
+#### Set the registrar
```
jacobheun: Let's delete this section, I think the property overview is sufficient for now. Also, the Registrar is going to set `topology.registrar` on the base Topology as well during registration, every Topology will have that property once they've registered, so it might be good to add that to the Topology

vasco-santos: Removed the section, added the `set` to the `Topology` and overwrite it in the `MulticodecTopology`

PR #16, opened 2019-12-20
fix: transport should not handle connection if upgradeInbound throws

vasco-santos:
Fixed per discussion on [libp2p/js-libp2p#523#discussion_r359997872](https://github.com/libp2p/js-libp2p/pull/523#discussion_r359997872)

### DIFF
--- a/src/transport/tests/listen-test.js
+++ b/src/transport/tests/listen-test.js
@@ -8,6 +8,7 @@ const expect = chai.expect
 chai.use(dirtyChai)
 const sinon = require('sinon')
 
+const pWaitFor = require('p-wait-for')
 const pipe = require('it-pipe')
 const { isValidTick } = require('./utils')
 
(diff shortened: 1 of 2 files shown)


src/transport/tests/listen-test.js
```diff
@@ -92,6 +92,23 @@ module.exports = (common) => {
...
+
+      const listener = transport.createListener(() => {
+        throw new Error('should not handle the connection if upgradeInbound throws')
+      })
+
+      // Listen
+      await listener.listen(addrs[0])
+
+      // Create a connection to the listener
+      const socket = await transport.dial(addrs[0])
```
jacobheun: @vasco-santos I think we should validate the socket was closed as well. Since the listener throws, it should hangup, so the dial socket should get closed. We could wait for `timeline.close` to exist after the dial before closing everything. `await pWaitFor(() => typeof socket.timeline.close === 'number')` or something similar.

Other than that, this looks good. I mainly just want to ensure we are properly closing the sockets on inbound errors.

vasco-santos: Yeah, I think that is better

PR #18, opened 2020-01-09
fix: dependencies for tests should not be needed by who requires the tests

vasco-santos:
Currently, transports must have `it-pipe` and `it-pair` in their dependencies, otherwise they are not able to run the interface tests


jacobheun:
It looks like `peer-info` is used by the exported Topology tests, can we move that into dependencies as well while we're at it?

vasco-santos:
you are right, adding it

PR #19, opened 2020-01-13
chore: peer discovery should emit peer id

vasco-santos:
BREAKING CHANGE: peer discovery event param is a PeerId instance instead of a PeerInfo instance

Context: https://github.com/libp2p/js-peer-info/issues/85

I added `_onPeer` as a mandatory function to be implemented, so that we can test the event emitted uniformly. Any better ideas to make it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "it-goodbye": "^2.0.1",
     "libp2p-tcp": "^0.14.1",
     "multiaddr": "^7.1.0",
+    "p-defer": "^3.0.0",
     "p-limit": "^2.2.1",
     "p-wait-for": "^3.1.0",
     "peer-id": "^0.13.3",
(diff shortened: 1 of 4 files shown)


jacobheun:
@vasco-santos I think we need to write up a clearer plan for how multiaddrs and protocols will make it into the PeerStore. The problem with this PR is that there isn't a clear way for multiaddrs for the peer to be tracked. Many of the discovery modules include the known addresses of the peer, if we change to this, we will lose that and won't be able to connect to the peers.

One approach might be to pass the PeerStore to the discovery modules. Instead of emitting peers they could put to the PeerStore, and add any other known metadata. We should probably flush out the v2 PeerStore before making changes here.

vasco-santos:
@jacobheun I agree, my idea would be to provide the `peerStore` so that the peers are properly tracked. Anyway, the discovery services should emit the `peer` event with the `peer-id` of a peer once it is discovered and added to the peerStore.

Anyway, we can start by having a clear definition for the `peerStorev2`

PR #34, opened 2020-02-13
fix: remove use of assert module

achingbrain:
The polyfill is big, we can simulate it by throwing an Error and it doesn't work under React Native.

### DIFF
--- a/src/connection/connection.js
+++ b/src/connection/connection.js
@@ -6,6 +6,52 @@ const withIs = require('class-is')
 const errCode = require('err-code')
 const Status = require('./status')
 
+function validateArgs (localAddr, localPeer, remotePeer, newStream, close, getStreams, stat) {
+  if (localAddr && !multiaddr.isMultiaddr(localAddr)) {
+    throw errCode(new Error('localAddr must be an instance of multiaddr'), 'ERR_INVALID_PARAMETERS')
+  }
+
+  if (!PeerId.isPeerId(localPeer)) {
+    throw errCode(new Error('localPeer must be an instance of peer-id'), 'ERR_INVALID_PARAMETERS')
+  }
+
+  if (!PeerId.isPeerId(remotePeer)) {
+    throw errCode(new Error('remotePeer must be an instance of peer-id'), 'ERR_INVALID_PARAMETERS')
+  }
+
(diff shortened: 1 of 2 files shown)


src/connection/connection.js
```diff
@@ -33,17 +30,49 @@ class Connection {
...
-    assert(PeerId.isPeerId(remotePeer), 'remotePeer must be an instance of peer-id')
-    assert(typeof newStream === 'function', 'new stream must be a function')
-    assert(typeof close === 'function', 'close must be a function')
-    assert(typeof getStreams === 'function', 'getStreams must be a function')
-    assert(stat, 'connection metadata object must be provided')
-    assert(stat.direction === 'inbound' || stat.direction === 'outbound', 'direction must be "inbound" or "outbound"')
-    assert(stat.timeline, 'connection timeline object must be provided in the stat object')
-    assert(stat.timeline.open, 'connection open timestamp must be provided')
-    assert(stat.timeline.upgraded, 'connection upgraded timestamp must be provided')
+    if (localAddr && !multiaddr.isMultiaddr(localAddr)) {
```
vasco-santos: what do you think on getting a util function to validate all those parameters?

otherwise, this adds so much noise to the constructor

achingbrain: I've pulled it out into a function.

Longer term using a validation library would be a much better option, or getting `libp2p` to validate things before passing them into constructors, or maybe not validating at all, given that we put little 'Transport compatible', 'Connection compatible', etc stickers all over our readmes (I'm kidding, we have tests for that).

If we go the validation library route, we should agree upfront which one to use so we don't have multiple modules that do the same job bloating bundle sizes.

achingbrain: I've pulled it out into a function.

Longer term using a validation library would be a much better option, or getting libp2p to validate things before passing them into constructors, or maybe not validating at all, given that we put little 'Transport compatible', 'Connection compatible', etc stickers all over our readmes (I'm kidding, we have tests for that).

If we go the validation library route, we should agree upfront which one to use so we don't have multiple modules that do the same job bloating bundle sizes.

achingbrain:
@jacobheun 

merge-pls2

PR #41, opened 2020-04-06
feat: peer-discovery not using peer-info

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR changes the `peer-discovery` interface to emit an object containing an `id` property and a `multiaddrs` property.

Future discovery services might gather other data, such as known protocols. However, we should specify those optional parameters when the time comes.

BREAKING CHANGE: peer-discovery emits object with id and multiaddrs properties


vasco-santos:
We currently do not have meaningful tests for this interface yet. Should we try to leverage this change to do it?

If we specify that the events should be emitted via a function `_onPeer(peerId, multiaddrs)`, we could call this function. I am not a big fan of this solution though. We could also have a base class that discovery services extend and it would do the `emit` for them.

cc @jacobheun

jacobheun:
> We currently do not have meaningful tests for this interface yet. Should we try to leverage this change to do it?

Yes we should.

> If we specify that the events should be emitted via a function _onPeer(peerId, multiaddrs), we could call this function. I am not a big fan of this solution though. We could also have a base class that discovery services extend and it would do the emit for them.

What are you trying to get out of this? It seems like it's doing some unnecessary restrictions. If we specify the API requirements and add tests that should be sufficient.

vasco-santos:
> What are you trying to get out of this? It seems like it's doing some unnecessary restrictions. If we specify the API requirements and add tests that should be sufficient.

I agree with not being nice to add unnecessary restrictions. However, we need to have a way of simulating the `peer` event from the discovery service and verify that it provided the `peerId` and `multiaddrs`. Any idea on how we should trigger it on a generic way?

jacobheun:
> Any idea on how we should trigger it on a generic way?

I think this should just be handled by the common setup requirements of the test. Such as, once `start` is called, the handler `discovery.on('peer', handler)` should be called repeatedly until `stop` is called. This would allow us to validate start, stop and whatever required event handler methods we add.

The interface tests shouldn't trigger anything, just call methods.

src/peer-discovery/README.md
```diff
@@ -71,4 +71,7 @@ It returns a `Promise`
 
 - `discovery.on('peer', (peerInfo) => {})`
 
-Everytime a peer is discovered by a discovery service, it emmits a `peer` event with the discover peer's [PeerInfo](https://github.com/libp2p/js-peer-info).
+Everytime a peer is discovered by a discovery service, it emmits a `peer` event with the discovered peer's information, which must contain the following properties:
```
jacobheun, line 74:
```suggestion
Every time a peer is discovered by a discovery service, it emits a `peer` event with the discovered peer's information, which must contain the following properties:
```

src/peer-discovery/README.md
```diff
@@ -71,4 +71,7 @@ It returns a `Promise`
...
+
+- `<`[`PeerId`](https://github.com/libp2p/js-peer-id)`>` `peerInfo.id`
+- `<Array<`[`Multiaddr`](https://github.com/multiformats/js-multiaddr)`>>` `peerInfo.multiaddrs`
```
jacobheun, line 77: Should we change `peerInfo` to `peerData` everywhere to avoid confusion?

src/peer-discovery/tests/index.js
```diff
@@ -30,5 +44,49 @@ module.exports = (common) => {
...
+        expect(id).to.exist()
+        expect(PeerId.isPeerId(id)).to.eql(true)
+        expect(multiaddrs).to.exist()
+
+        multiaddrs.forEach((m) => expect(multiaddr.isMultiaddr(m)).to.eql(true))
+
+        defer.resolve()
+      })
+
+      return defer.promise
```
jacobheun, line 62: Avoid mixing async and promise usage

```suggestion
      await defer.promise
```

src/peer-discovery/tests/index.js
```diff
@@ -30,5 +44,49 @@ module.exports = (common) => {
       await discovery.start()
       await discovery.start()
     })
+
+    it('should listen a peer event after start', async () => {
```
jacobheun, line 48: emit not listen?
```suggestion
    it('should emit a peer event after start', async () => {
```

src/peer-discovery/tests/index.js
```diff
@@ -1,17 +1,31 @@
...
+const delay = require('delay')
+const pDefer = require('p-defer')
+
 module.exports = (common) => {
   describe('interface-peer-discovery', () => {
     let discovery
 
-    before(async () => {
+    beforeEach(async () => {
       discovery = await common.setup()
```
jacobheun, line 19: If `common.setup()` needs to be run before every test, we need to tear it down after every test. The compliance test is causing this to leak intervals.

test/peer-discovery/compliance.spec.js
```diff
@@ -5,12 +5,22 @@ const tests = require('../../src/peer-discovery/tests')
...
+        discoveryDelay: 1
+      })
+
+      intervalId = setInterval(mockDiscovery._discoverPeer, 1000)
+
+      return mockDiscovery
     },
     async teardown () {
+      clearInterval(intervalId)
       await new Promise(resolve => setTimeout(resolve, 10))
```
jacobheun, line 24: This shouldn't need to be here.
```suggestion
```

src/peer-discovery/README.md
vasco-santos: yes please 😂

vasco-santos:
> Found a few things. Also, I think we should merge this into a new minor branch since we'll need to update a few of the other interfaces for the peer-info change.

What about an approach with a pre-release, so that we can start merging all the PRs related to `peer-discovery` as release candidates, and then we only need to change for the final release? All the needed PRs are open now as well, so I think the sooner we get them, the sooner we unblock other PRs

jacobheun:
> What about an approach with a pre-release, so that we can start merging all the PRs related to peer-discovery as release candidates, and then we only need to change for the final release? 

These aren't mutually exclusive options. We can do pre releases / RCs from any branch. The main reason for the separate branch is to keep master open for any patches that might be needed before this is released.

We can always cherry pick patches onto the latest tag and release that, but we'd also have to remember to pull those patches, as needed back into master. Putting everything into a new branch avoids the need to remember because we'll get the change when we go to merge into master.

jacobheun:
This also replaces https://github.com/libp2p/js-libp2p-interfaces/pull/19 right?

vasco-santos:
> This also replaces #19 right?

Oh yes! Just closed it, completely forgot about that initial PR

PR #42, opened 2020-04-07
chore: remove peer-info usage on topology

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR removes the `peer-info` usage on topology.

This changes have impact on pubsub modules, as well as the dht

BREAKING CHANGE: topology api now uses peer-id instead of peer-info

### DIFF
--- a/src/topology/README.md
+++ b/src/topology/README.md
@@ -56,8 +56,8 @@ const toplogy = new MulticodecTopology({
   max: 50,
   multicodecs: ['/echo/1.0.0'],
   handlers: {
-    onConnect: (peerInfo, conn) => {},
-    onDisconnect: (peerInfo) => {}
+    onConnect: (peerId, conn) => {},
+    onDisconnect: (peerId) => {}
   }
 })
 ```
@@ -69,8 +69,8 @@ The `MulticodecTopology` extends the `Topology`, which makes the `Topology` API 
 ###  Topology
 
 - `Topology`
-  - `peers<Map<string, PeerInfo>>`: A Map of peers belonging to the topology.
-  - `disconnect<function(PeerInfo)>`: Called when a peer has been disconnected
+  - `peers<Map<string, PeerId>>`: A Map of peers belonging to the topology.
+  - `disconnect<function(PeerId)>`: Called when a peer has been disconnected
 
 #### Constructor
 
@@ -79,8 +79,8 @@ const toplogy = new Topology({
   min: 0,
   max: 50,
   handlers: {
-    onConnect: (peerInfo, conn) => {},
-    onDisconnect: (peerInfo) => {}
+    onConnect: (peerId, conn) => {},
+    onDisconnect: (peerId) => {}
   }
 })
 ```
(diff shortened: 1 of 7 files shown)


src/topology/README.md
```diff
@@ -95,27 +95,27 @@ const toplogy = new Topology({
...
+- `peerId` is the [PeerId][peer-id] of the peer to add.
 
 #### Notify about a peer disconnected event
 
-- `topology.disconnect(peerInfo)`
+- `topology.disconnect(peerId)`
 
 **Parameters**
-- `peerInfo` is the [PeerInfo][peer-info] of the peer disconnected.
+- `peerId` is the [PeerIid][peer-id] of the peer disconnected.
```
jacobheun, line 111:
```suggestion
- `peerId` is the [PeerId][peer-id] of the peer disconnected.
```

jacobheun: We're apparently missing the ref link for these at the bottom of the readme, we need to add that.

src/topology/multicodec-topology.js
```diff
@@ -55,43 +55,44 @@ class MulticodecTopology extends Topology {
 
   /**
    * Update topology.
-   * @param {Array<PeerInfo>} peerInfoIterable
+   * @param {Array<PeerData>} peerDataIterable
```
jacobheun, line 58: Where is `PeerData` declared? We need a type def, or just document the literal object here.
```suggestion
   * @param {Array<{id: PeerId, multiaddrs: Array<Multiaddr>}>} peerDataIterable
```

src/topology/index.js
```diff
@@ -26,6 +26,7 @@ class Topology {
     this._onConnect = handlers.onConnect || noop
     this._onDisconnect = handlers.onDisconnect || noop
 
+    // TODO: can this be a set?
```
jacobheun, line 29: The issue with making this a Set is that we can't test equality of complex objects. So if we end up with a copy of an existing peer it will get added to the set. The Map gets around this by using the peer id string as the key.

src/topology/multicodec-topology.js
```diff
@@ -55,43 +55,44 @@ class MulticodecTopology extends Topology {
 
   /**
    * Update topology.
-   * @param {Array<PeerInfo>} peerInfoIterable
+   * @param {Array<{id: PeerId, multiaddrs: Array<Multiaddr>}>} peerDataIterable
```
jacobheun, line 58: I missed protocols in the last suggestion which is the one we care about
```suggestion
   * @param {Array<{id: PeerId, multiaddrs: Array<Multiaddr>, protocols: Array<string>}>} peerDataIterable
```

src/topology/multicodec-topology.js
```diff
@@ -55,43 +55,44 @@ class MulticodecTopology extends Topology {
...
    * @returns {void}
    */
-  _updatePeers (peerInfoIterable) {
-    for (const peerInfo of peerInfoIterable) {
-      if (this.multicodecs.filter(multicodec => peerInfo.protocols.has(multicodec)).length) {
+  _updatePeers (peerDataIterable) {
+    for (const peerData of peerDataIterable) {
```
jacobheun, line 62: nit: exploding the properties might tidy this function up a little.
```js
for (const { id, protocols }  of peerDataIterable) {
```

vasco-santos: done :)

PR #44, opened 2020-04-10
chore: add discv5 peer discovery module

wemeetagain:


vasco-santos:
@wemeetagain 

FYI in `libp2p@0.28` we will change the `peer-discovery` module. It will probably change as https://github.com/libp2p/js-libp2p-interfaces/pull/41 but I will let you know. By that time, it will be useful to add the compliance tests, since they are updated there.

I will aldo add this module to the `js-libp2p` packages table

vasco-santos:
@wemeetagain can we have `discv5` with a leadMaintainer in the `package.json`?

For https://github.com/libp2p/js-libp2p/pull/606, we use a tool which gathers this data from the repos to add the leadMaintainer in the `js-libp2p` Readme package table.

wemeetagain:
@vasco-santos :+1: We're looking forward to 0.28, deprecated `PeerInfo`, new `PeerStore`, etc
Let us know when would be appropriate to revisit upgrading.

We just updated discv5 `package.json` with the `leadMaintainer`.

PR #52, opened 2020-06-16
feat: record interface

vasco-santos:
This PR adds the `record-interface`.

The `record-interface` primary goal is to guarantee that any record implementation will be compliant with the expectations of a libp2p **envelope**.

`js-libp2p` will start by having a single implementation of the `record-interface`, known as a `peer-record`. This implementations will be used to share peers' multiaddrs across the network with security guarantees, thanks to the libp2p envelope where the records will be wrapped into.

Open questions:
- Should we have tests for the unmarshal operation in the interface tests? I think that we should have the unmarshal and then rely on `isEqual` to verify if the unmarshal operation resulted in the same content as before the marshal

A record implementation will now look as follows:
```js
class PeerRecord extends Record {
  // ....
}

PeerRecord.createFromProtobuf = (buf) => {
  // ...
}

module.exports = PeerRecord
```
The `createFromProtobuf` will return an instance of the Peer Record.

References:
- https://github.com/libp2p/specs/pull/217
- https://github.com/libp2p/go-libp2p-core/blob/master/record/record.go

### DIFF
--- a/src/record/README.md
+++ b/src/record/README.md
@@ -1,0 +1,75 @@
+interface-record
+==================
+
+A libp2p node needs to store data in a public location (e.g. a DHT), or rely on potentially untrustworthy intermediaries to relay information. Libp2p provides an all-purpose data container called **envelope**, which includes a signature of the data, so that it its authenticity can be verified.
+
+The record represents the data that will be stored inside the **envelope** when distributing records across the network. The `interface-record` aims to guarantee that any type of record created is compliant with the libp2p **envelope**.
+
+Taking into account that a record might be used in different contexts, an **envelope** signature made for a specific purpose **must not** be considered valid for a different purpose. Accordingly, each record has a short and descriptive string representing the record use case, known as **domain**. The data to be signed will be prepended with the domain string, in order to create a domain signature.
+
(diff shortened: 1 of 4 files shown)


src/record/README.md
```diff
@@ -0,0 +1,75 @@
...
+class PeerRecord extends Record {
+  constructor (peerId, multiaddrs, seqNumber) {
+    super (ENVELOPE_DOMAIN_PEER_RECORD, ENVELOPE_PAYLOAD_TYPE_PEER_RECORD)
+  }
+
+  marshal () {
+    // Implement and return using Protobuf
+  }
+
+  isEqual (other) {
```
jacobheun, line 47: As mentioned in https://github.com/libp2p/js-libp2p/pull/681, I think this should be `equals`.

vasco-santos: It makes sense, and `peer-id` is deprecating `isEqual` to use `equals`. But, somehow I did not bring that consistency here 🤦‍♂️ 
Thanks for flagging it.

vasco-santos:
This should be ready to get merged in.
@jacobheun would you prefer this to get merged on a `0.3` branch before rolling all the releases?

jacobheun:
> would you prefer this to get merged on a 0.3 branch 

master is 0.3 and I think we're fine to move there, this is backwards compatible.

vasco-santos:
lol yes, github UI shows the latest release in the git repo and not the latest 🤦‍♂️ 

image

PR #54, opened 2020-07-03
fix: reconnect should trigger topology on connect if protocol stored

vasco-santos:
Considering a scenario where Peer A and B start, connect to each other using a multicodec topology. Then, Peer B restarts, it is still not connected to A. It will run the update peers: https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/topology/multicodec-topology.js#L68 but it does not have a connection to A yet, so it will not call onConnect. After this, identify will run, but the protocols will not change and the protocol change handler will not be called.

Fixes https://github.com/libp2p/js-libp2p/issues/658

### DIFF
--- a/src/topology/multicodec-topology.js
+++ b/src/topology/multicodec-topology.js
@@ -113,7 +113,7 @@ class MulticodecTopology extends Topology {
       return
     }
 
-    if (this.multicodecs.filter(multicodec => protocols.includes(multicodec)).length) {
+    if (this.multicodecs.find(multicodec => protocols.includes(multicodec))) {
       this.peers.add(peerId.toB58String())
       this._onConnect(peerId, connection)
     }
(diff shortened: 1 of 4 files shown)


src/topology/multicodec-topology.js
```diff
@@ -97,6 +99,25 @@ class MulticodecTopology extends Topology {
...
+   */
+  _onPeerConnect (connection) {
+    const peerId = connection.remotePeer
+    const protocols = this._registrar.peerStore.protoBook.get(peerId)
+
+    if (!protocols) {
+      return
+    }
+
+    if (this.multicodecs.filter(multicodec => protocols.includes(multicodec)).length) {
```
jacobheun, line 116: nit: I'd change this to a find, we don't need to go through all the multicodecs if we find one.

jacobheun:
We may want to backport this to 0.2.x for those who still need to upgrade from the peerinfo changes.

vasco-santos:
On it

vasco-santos:
It is not as easy as this one. Registrar does not have access to the `connectionManager` events, as in `libp2p@0.27` we have the connect events in `libp2p`: https://github.com/libp2p/js-libp2p/blob/0.27.x/src/registrar.js

I need to add libp2p to registrar in `0.27.x` branch and use it here in `0.2.x`. Is something we should do?

PR #55, opened 2020-07-15
chore: update deps

jacobheun:
Most notably streaming-iterables from 4.x to 5.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "p-limit": "^2.3.0",
     "p-wait-for": "^3.1.0",
     "peer-id": "^0.13.13",
-    "peer-info": "^0.17.5",
     "sinon": "^9.0.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -45,19 +45,19 @@
...
+    "it-pipe": "^1.1.0",
+    "libp2p-tcp": "^0.14.5",
+    "multiaddr": "^7.5.0",
     "p-defer": "^3.0.0",
     "p-limit": "^2.3.0",
     "p-wait-for": "^3.1.0",
-    "peer-id": "^0.13.11",
-    "peer-info": "^0.17.0",
+    "peer-id": "^0.13.13",
+    "peer-info": "^0.17.5",
```
vasco-santos, line 55:
```suggestion
```

ups, this should just go away

Issue #53, opened 2020-07-02
RFC: pubsub-interface


vasco-santos:
## Background

We created [libp2p-pubsub](https://github.com/libp2p/js-libp2p-pubsub) as the base implementation of a pubsub router when we were working on the second pubsub router for libp2p (gossipsub).

While this approach allowed us to guarantee the implementation of pubsub routers with the same external APIs, we still have a lot of duplication in the tests. This comes with an extra cost for code maintenance. Over time, it is expected that new pubsub router implementations might appear.

## Proposal

With gossipsub1.1 update, @wemeetagain is doing some improvements on the base pubsub implementation that will be moved to the repo. With this refactor, I think that we have a clear implementation and API for our base pubsub.

My proposal is to create a `pubsub-interface` and move the `libp2p-pubsub` codebase into this repo. Moreover, this repo should provide a test suite that each pubsub router should run to have an extensive test of its external API. The routers would only need to test their internal behaviours and require the interface tests for the rest.

After landing the gossipsub1.1 it will be a good opportunity for doing this. We will need to update floodsub to receive `libp2p` (as we are doing for gossipsub), which will require changes in the tests. I plan to get it done in parallel with this effort.

vasco-santos:
Thoughts?
cc @jacobheun @wemeetagain

vasco-santos:
This was done

PR #60, opened 2020-08-16
feat: interface pubsub

vasco-santos:
This PR adds pubsub-interface, which will replace [libp2p/js-libp2p-pubsub](https://github.com/libp2p/js-libp2p-pubsub) per https://github.com/libp2p/js-libp2p-interfaces/issues/53 . This is also an important work in the context of dropping the pubsub abstraction layer in `js-libp2p` per https://github.com/libp2p/js-libp2p-pubsub/issues/68

The interface is closely in line with the [base pubsub spec](https://github.com/libp2p/specs/tree/master/pubsub) and also got some inspiration from `go-libp2p-pubsub` during the gossipsub refactor in preparation for 1.1.

This PR includes:
- move `libp2p-pubsub` content into `src/pubsub` and `test/pubsub`.
- refactor of `libp2p-pubsub` codebase regarding constructor and to better track streams
- test suite that pubsub routers should use for testing

TODO:

- [x] routers test suite
- [x] final refactor on pending libp2p-gossipsub
- [x] base implementation test suite refactor
- [x] improve docs on proper usage, what to override and what not to override

Unblocks:

- [ ] https://github.com/libp2p/js-libp2p-floodsub/pull/108
- [ ] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/122


vasco-santos:
@wemeetagain can you give a pass through this PR?
It is gigantic, but it is mostly for moving pieces around, both from `libp2p-pubsub` and your refactor from gossipsub. The main changes I needed to do are in the constructor, subscription handlers and some improvements on the coordination of overrides.

I did these changes in parallel with the routers update https://github.com/libp2p/js-libp2p-floodsub/pull/108 and https://github.com/ChainSafe/js-libp2p-gossipsub/pull/122 which are successfully using this code now

src/pubsub/index.js
```diff
@@ -0,0 +1,689 @@
...
+    return utils.msgId(msg.from, msg.seqno)
+  }
+
+  /**
+   * Process publish message received from a peer
+   * @abstract
+   * @param {InMessage} msg
+   * @returns {void}
+   */
+  _publishFrom (msg) {
```
vasco-santos, line 439: I need to remove this according to the gossipsub last PR merged!

src/pubsub/README.md
```diff
@@ -0,0 +1,218 @@
...
+Subscribe to the given topic(s).
+
+#### `pubsub.subscribe(topics, [handler])`
+
+##### Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| topics | `Array<string>|string` | set of pubsub topics |
+| [handler] | `function (msg)` | handler for messages received in the given topics |
```
wemeetagain, line 121: I think the interface should only allow subscription/unsubscription of a single topic at a time.

Attaching a handler to multiple topics at once feels like an anti-pattern, or at least a less primary interface, in the same way we were previously attaching multiple topics to multiple messages at once in `publish`.

I think this will simplify some of the underlying implementations, too. The body of our subscription code looks roughly like: `for (const topic of topics) { /* doSubscribe(topic) */ }`. We would be able to remove this outer loop and better support extended behavior.

Also, go-libp2p-pubsub's interface for subscribing is for a single topic at a time, for what its worth.

src/pubsub/index.js
```diff
@@ -0,0 +1,689 @@
...
+
+  /**
+   * Overriding the implementation of _encodeRpc should use the appropriate router protobuf.
+   * Encode RPC object into a Uint8Array.
+   * @abstract
+   * @param {RPC} rpc
+   * @returns {Uint8Array}
+   */
+  _encodeRpc (rpc) {
+    throw errcode(new Error('_encodeRpc must be implemented by the subclass'), 'ERR_NOT_IMPLEMENTED')
```
wemeetagain, line 473: In the same way that `_acceptFrom` returns a default implementation, we should consider providing default encoding/decoding using the base protobuf.

src/pubsub/index.js
```diff
@@ -0,0 +1,689 @@
...
+
+    const from = this.peerId.toB58String()
+    const msgObject = {
+      receivedFrom: from,
+      from: from,
+      data: message,
+      seqno: utils.randomSeqno(),
+      topicIDs: topics
+    }
+
```
wemeetagain, line 591: pls copy the latest edits from gossipsub here, so we can have the signature already attached in `_publish`

src/pubsub/index.js
```diff
@@ -0,0 +1,689 @@
...
+  }
+
+  /**
+   * Subscribes to topics.
+   * @abstract
+   * @param {Array<string>|string} topics
+   * @param {function} [handler]
+   * @returns {void}
+   */
+  subscribe (topics, handler) {
```
wemeetagain, line 618: Previously, we had this broken into two levels `subscribe` and `_subscribe`.
Perhaps this can be better named, eg: `subscribe` and `_newSubscription` (or some other).
But the outer layer, `subscribe` normalized input and removed existing subscriptions, so that `_newSubscription` was able to be extended in a way that didn't need to renormalize + remove existing subscriptions, but rather acted on known new subscriptions.
Eg: In gossipsub, we want to call `join` only on new subscriptions, so overriding just using this method becomes somewhat more cumbersome, than overriding a lower level method for "subscribing to new, previously unsubscribed topic".

src/pubsub/README.md
vasco-santos: I agree that the API would look better with one topic. Taking into account that we are doing several breaking changes in pubsub, I think we should also change this.

The libp2p API doc also says a [single](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#pubsubsubscribe) topic, so we were supporting it below in the stack but we were not mentioning it. However, this should be pointed in the migration guide

src/pubsub/index.js
vasco-santos: I went with not decoupling it because all these operations seem mandatory to do. For the gossipsub example, I changed it to:

```js
subscribe (topics: string[], handler: (msg: any) => void): void {
    super.subscribe(topics, handler)
    this.join(topics)
  }
```

Any reason you foresee to not simplify this?

src/pubsub/README.md
vasco-santos: What about publish? Should we also just accept one topic per API call? That's how we have it in libp2p main docs as well

src/pubsub/index.js
wemeetagain: I think this is fine, especially considering we'll only be operating on a single topic.

src/pubsub/index.js
wemeetagain: It looks like this wasn't copied exactly, there was a reason it was as it was before:
- we want to emit the message object with the optional signature (since other messages will be emitted with the signature and we shouldn't be treating self-published messages differently)
- _publish currently expects an "InMessage" message transformed using `utils.normalizeRpcInMessage`. It should be standardized bc this function is called in different contexts: self-publish and peer-publish. Also, all overridable methods that operate on messages (_publish, _processRpcMessage, etc) receive an "InMessage". It's nice to rely on that and not require overriders to call 'utils.normalizeRpcInMessage'.

src/pubsub/index.js
```diff
@@ -0,0 +1,663 @@
...
+   */
+  _publish (message) {
+    throw errcode(new Error('publish must be implemented by the subclass'), 'ERR_NOT_IMPLEMENTED')
+  }
+
+  /**
+   * Subscribes to a given topic.
+   * @abstract
+   * @param {string} topic
+   * @param {function} [handler]
```
wemeetagain, line 606: Is there somewhere where we can document what the input of this function will be? should be `(msg: InMessage) => void`

src/pubsub/index.js
```diff
@@ -0,0 +1,663 @@
...
+      throw new Error('FloodSub is not started')
+    }
+
+    if (!handler) {
+      this.removeAllListeners(topic)
+    } else {
+      this.removeListener(topic, handler)
+    }
+
+    if (this.listenerCount(topic) === 0) {
```
wemeetagain, line 641: What is this for?
I thought we would want to delete the topic from `this.subscriptions` no matter what. As well as send unsubscriptions.

src/pubsub/index.js
vasco-santos: We had a test before stating: "should emit non normalized messages on publish" and that's why I put it like this.

I will revisit this in a bit

src/pubsub/index.js
vasco-santos: Imagine you have a webapp that subscribes a specific topic in a UI component and the same topic in a different UI component, both with a handler. If one unmounts, the topic would be unsubscribed and no more messages would be received.
I think it makes sense when people use handlers. If you do not provide an handler, the subscription will be always removed.

src/pubsub/index.js
vasco-santos: good call! I can document it here

src/pubsub/README.md
vasco-santos: I will update it

src/pubsub/index.js
wemeetagain: Yeah, the code is a little screwy, I think a proper solution would be to allow an InMessage to be signed (and have signature/key attached).

What we were doing to get around all this was:
- create msg: InMessage
- normalize/sign `outMsg = await this.buildMessage(msg)`
- set msg to the non-normalized msg, but including the signature (`msg = utils.normalizeInRpcMsg(outMsg)`)
- _then_ emit msg
- _publish(msg)

vasco-santos: I guess we can emit it normalized and signed for now. I don't see any reason against it. I changed to what you had before for now

vasco-santos:
> It would be nice for typescript router implementers to have a typescript interface

Ideally, we should move forward with the support via js doc + aegir soon

src/pubsub/index.js
wemeetagain: I see. IMO we should reconsider this interface, or at least provide a separate subscription-only interface. Topic subscription should have the ability to be separated from topic event handling, but combining the two here seems to force this coupling, even if its not desired.

I think the bad behavior is removing all listeners on unsubscribe. Because of this, consumers _must_ couple event handling and subscription, because there is no way to perform unsubscription without editing attached event handlers.

vasco-santos: Well, I have been thinking about it. now that we expose the router directly by libp2p.pubsub , we can probably just leverage the event emitter api, document that the topics are even emitter events and people just need to use its api. It is a new breaking change, but I think we are in a good position to so it!
We definitely need to have a good migration guide for this

vasco-santos: We just cannot accept the subscription change event name as a topic.

vasco-santos: `js-ipfs` also relies on the `handlers`. @jacobheun what are your thoughts on getting rid of them?

src/pubsub/message/sign.js
```diff
@@ -0,0 +1,92 @@
...
+ * Verifies the signature of the given message
+ * @param {InMessage} message
+ * @returns {Promise<Boolean>}
+ */
+async function verifySignature (message) {
+  // Get message sans the signature
+  const baseMessage = { ...message }
+  delete baseMessage.signature
+  delete baseMessage.key
+  if (typeof baseMessage.from === 'string') {
```
jacobheun, line 42: Won't this always be a string as this takes an `InMessage`? The message is run through the normalizeIn function by this point. If this is not a string there's an issue.

src/pubsub/message/sign.js
```diff
@@ -0,0 +1,92 @@
...
+ * @returns {Promise<PublicKey>}
+ */
+async function messagePublicKey (message) {
+  // should be available in the from property of the message (peer id)
+  let from
+  if (typeof message.from === 'string') {
+    from = PeerId.createFromB58String(message.from)
+  } else {
+    from = PeerId.createFromBytes(message.from)
+  }
```
jacobheun, line 71: `PeerId.createFromCID` will handle both these cases.

```suggestion
  const from = PeerId.createFromCID(message.from)
```

src/pubsub/index.js
```diff
@@ -0,0 +1,663 @@
...
+      return false
+    }
+
+    if (msgs.length) {
+      msgs.forEach(message => {
+        if (!(this.canRelayMessage || message.topicIDs.some((topic) => this.subscriptions.has(topic)))) {
+          this.log('received message we didn\'t subscribe to. Dropping.')
+          return
+        }
+        const msg = utils.normalizeInRpcMessage(message, PeerId.createFromB58String(idB58Str))
```
jacobheun, line 357: The PeerId passed to `normalizeInRpcMessage` is converted to its string form for use, and I only see this called with a PeerId here. Perhaps this should just change to taking and using the string to avoid the double conversion.

src/pubsub/index.js
```diff
@@ -0,0 +1,663 @@
...
+        this._processRpcMessage(msg)
+      })
+    }
+    return true
+  }
+
+  /**
+   * Handles an subscription change from a peer
```
jacobheun, line 365:
```suggestion
   * Handles a subscription change from a peer
```

src/pubsub/index.js
```diff
@@ -0,0 +1,663 @@
...
+
+  /**
+   * Send an rpc object to a peer
+   * @param {string} id peer id
+   * @param {RPC} rpc
+   * @returns {void}
+   */
+  _sendRpc (id, rpc) {
+    const peerStreams = this.peers.get(id)
+    if (!peerStreams || !peerStreams.isWritable) {
```
jacobheun, line 472: This should be an error, we need to log this at the very least, I'd prefer a throw I think. We shouldn't be trying to send data to peers we don't have streams to.

jacobheun (commented):
Just a few minor things left

test/pubsub/topic-validators.spec.js
```diff
@@ -0,0 +1,111 @@
...
+        data: uint8ArrayFromString('a different message'),
+        seqno: utils.randomSeqno(),
+        topicIDs: [filteredTopic]
+      }]
+    }
+
+    // process previously invalid message, now is valid
+    pubsub._processRpc(peer.id.toB58String(), peer, invalidRpc2)
+    pubsub.unsubscribe(filteredTopic)
+    await delay(500)
```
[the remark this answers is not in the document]
vasco-santos: That is true! The problem with these delays is that the `p-wait-for` would need to go into router specificity. In this case, we would need to wait for the heartbeat in gossipsub, while floodsub does not need the delay. THat's why I used the delay, I could not find a cleaner approach. The better way I thought was to provide in the tests setup a way of the router to provide a function to await. If this becomes a problem with flakey tests, we can figure out a better I guess. What do you think?

src/pubsub/tests/multiple-nodes.js
```diff
@@ -0,0 +1,346 @@
...
+              defer.resolve()
+            }
+          }
+
+          return defer.promise
+        })
+      })
+    })
+
+    describe('only some nodes subscribe the networks', () => {
```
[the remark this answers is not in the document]
vasco-santos: They are a copy from the old tests we had in `libp2p-floodsub` and `libp2p-pubsub`. Can I open an issue and tackle this as a follow up?

test/pubsub/topic-validators.spec.js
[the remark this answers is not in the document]
jacobheun: It might be needed for the compliance tests but this particular file is for the pubsub interface itself, not downstream implementations

src/pubsub/tests/multiple-nodes.js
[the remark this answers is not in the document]
jacobheun: yeah that works

test/pubsub/topic-validators.spec.js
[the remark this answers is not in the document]
vasco-santos: removed the delays for the topic validators. I did not notice this review comment was regarding them

src/pubsub/tests/multiple-nodes.js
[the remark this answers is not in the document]
vasco-santos: https://github.com/libp2p/js-libp2p-interfaces/issues/63

jacobheun (approved):
LGTM

PR #62, opened 2020-08-23
fix: replace remaining Buffer usage with Uint8Array

jacobheun:
This replaces the remaining usages of Buffer with Uint8Array.

BREAKING CHANGE: this changes the API of records and updates several tests to use Uint8Arrays

required by https://github.com/libp2p/js-libp2p/pull/730


achingbrain (approved):

vasco-santos (approved):
LGTM

PR #64, opened 2020-09-04
Add type generation from jsdoc

mpetrunic:
Related to: https://github.com/libp2p/js-libp2p/issues/659

I've fixed some jsdocs to generate better types. There is also some issue with index modules but overall jsdoc seems to be
generating quite good types. A lot of stuff related to protons library is "any" since that library doesn't have types. Not sure from where to import/use all those stream/duplex/DuplexIterable types

Do we wan't some precommit hook to regenerate types?

cc @wemeetagain @tuyennhv you can try from this branch directly


index.d.ts
```diff
@@ -0,0 +1,769 @@
...
+    export function signMessage(peerId: import("peer-id"), message: any): Promise<any>;
+    export const SignPrefix: any;
+    /**
+     * Verifies the signature of the given message
+     * @param {InMessage} message
+     * @returns {Promise<Boolean>}
+     */
+    export function verifySignature(message: any): Promise<boolean>;
+}
+declare module "pubsub/index" {
```
tuyennhv, line 432: the error I got from gossipsub is that `Could not find a declaration file for module 'libp2p-interfaces/src/pubsub'`. Does this mean gossipsub have to import module "pubsub/index" instead?

mpetrunic: I think i need to change rootDir 🤨

mpetrunic:
cc @vasco-santos is this proper way to include types?

vasco-santos:
@mpetrunic The goal is to not have the `index.d.ts` in the repo. `aegir` should generate it from the JS docs when we release each module. This way, it is always updated with the current jsdoc. However, it seems that `aegir` did not release yet the support for it https://github.com/ipfs/aegir/pull/568

@hugomrdias do you have any visibility on when you will be able to go back to the aegir support?

vasco-santos:
If we do not have support soon for `aegir`, we can move forward with this as the migration should be as simple as removing the declaration file from the repo and `aegir` generate it. @mpetrunic let us know once we should review this PR

hugomrdias:
> @mpetrunic The goal is to not have the `index.d.ts` in the repo. `aegir` should generate it from the JS docs when we release each module. This way, it is always updated with the current jsdoc. However, it seems that `aegir` did not release yet the support for it [ipfs/aegir#568](https://github.com/ipfs/aegir/pull/568)
> 
> @hugomrdias do you have any visibility on when you will be able to go back to the aegir support?

no i don't, it will take a while probably

mpetrunic:
@vasco-santos 
Seems like we can export single type declaration only if we export everything from src/index.js file and use that only.

@wemeetagain @tuyennhv 
Imports work now, but gossipsub has a lot of duplicated/incompatible types with libp2p-interfaces. We should probably try to port PeerStreams type definition into jsdoc here.

wemeetagain:
> gossipsub has a lot of duplicated/incompatible types with libp2p-interfaces

:+1: the types there are meant to be temporary until these types (and js-libp2p core) types are available. Happy to slowly get rid of these libp2p types duplicated everywhere :joy: 
@tuyennhv has a branch https://github.com/ChainSafe/js-libp2p-gossipsub/compare/tuyen/libp2p-interfaces-types-2 where we can sort this out.

vasco-santos:
> Seems like we can export single type declaration only if we export everything from src/index.js file and use that only. So I changed tsconfig to generate type declaration alongside js sources.

Yeah, that should be because we require specific interfaces

hugomrdias:
@gozala has taken over the ts jsdocs aegir PRs and is making good progress. Maybe that helps here.

mpetrunic:
> > Seems like we can export single type declaration only if we export everything from src/index.js file and use that only. So I changed tsconfig to generate type declaration alongside js sources.
> 
> Yeah, that should be because we require specific interfaces
> 
> @mpetrunic should I review the PR or should we wait on gossipsub integration before?

@vasco-santos I'm still getting types inline with gossipsub and removing duplicates, so pls wait a bit more before reviewing.

> @Gozala has taken over the ts jsdocs aegir PRs and is making good progress. Maybe that helps here.

@hugomrdias We can switch over once it's ready, I'm mostly changing jsdoc to describe types more accurately.

src/pubsub/index.js
```diff
@@ -91,7 +105,7 @@ class PubsubBaseProtocol extends EventEmitter {
     /**
      * Map of peer streams
      *
-     * @type {Map<string, PeerStreams>}
+     * @type {Map<string, import('./peer-streams')>}
```
tuyennhv, line 108: can you also change the below topic validator function from
```
@typedef {function(string, RPC): boolean} validator
```
to 
```
@typedef {function(string, InMessage): void} validator
```

src/pubsub/index.js
```diff
@@ -118,7 +132,7 @@ class PubsubBaseProtocol extends EventEmitter {
 
     /**
      * Topic validator function
-     * @typedef {function(string, RPC): boolean} validator
+     * @typedef {function(string, InMessage): boolean} validator
```
tuyennhv, line 135: should return `void` instead

src/pubsub/index.js
```diff
@@ -118,7 +132,7 @@ class PubsubBaseProtocol extends EventEmitter {
 
     /**
      * Topic validator function
-     * @typedef {function(string, RPC): boolean} validator
+     * @typedef {function(string, InMessage): void} validator
```
tuyennhv, line 135: should return `Promise<void>` instead of `void` :)

tuyennhv:
Confirm that gossipsub works fine with this PR 👍

mpetrunic:
@vasco-santos we are ready for review :)

src/connection/connection.js
```diff
@@ -133,21 +133,22 @@ class Connection {
...
      * User provided tags
+     * @type {string[]}
      */
     this.tags = []
   }
 
   /**
    * Get connection metadata
-   * @return {Object}
+   * @this {Connection}
```
vasco-santos, line 143: Should we document the object returned here?
`{ direction: string, ... }`

src/connection/connection.js
```diff
@@ -133,21 +133,22 @@ class Connection {
...
    */
   get stat () {
     return this._stat
   }
 
   /**
    * Get all the streams of the muxer.
-   * @return {Array<*>}
+   * @this {Connection}
```
vasco-santos, line 151: Why this change?

vasco-santos (commented):
Overall this looks good to me. Just a small comments below

src/connection/connection.js
mpetrunic: I've hinted that context of this is Connection so it will automatically pickup type of this._getStreams().

package.json
```diff
@@ -4,16 +4,21 @@
   "description": "Interfaces for JS Libp2p",
   "leadMaintainer": "Jacob Heun <jacobheun@gmail.com>",
   "main": "src/index.js",
+  "types": "types/src/index.d.ts",
```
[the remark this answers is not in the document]
mpetrunic: Well since there is nothing exported from index, we don't need it yeah

src/connection/connection.js
mpetrunic: Since I've hinted context, it picked up those things automatically:
https://github.com/libp2p/js-libp2p-interfaces/pull/64/files#diff-3ca9b26cdc518eaea397963f76e33cd5R102

vasco-santos: ok, cool! sounds good!

package.json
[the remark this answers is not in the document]
vasco-santos: ok, so other than removing this, I think we are good

wemeetagain:
I think this is ready for re-review

vasco-santos (approved):
LGTM! 🎉

vasco-santos:
FYI Shipped `libp2p-interfaces@0.5.2`

PR #65, opened 2020-10-01
feat: update pubsub getMsgId return type to Uint8Array

wemeetagain:
See https://github.com/libp2p/specs/pull/285

BREAKING CHANGE:
new getMsgId return type is not backwards compatible with prior `string`
return type.


vasco-santos:
Thanks for updating this @wemeetagain 
Can you fix the lint fail?

vasco-santos (approved):

PR #66, opened 2020-10-11
feat: pubsub: add global signature policy

wemeetagain:
See https://github.com/libp2p/specs/pull/294 and https://github.com/libp2p/specs/pull/299

Add a "global" signature policy to pubsub. This is a first-step towards having a "topic-specific" signature policy as described in the specs.
This "signature policy" replaces the ad-hoc options `signMessages` and `strictSigning` that were previously in use.

Only 'StrictSign' and 'StrictNoSign' signature policies are implemented, the 'LaxSign' and 'LaxNoSign' signature policies are not recommended and will be deprecated, so are not included.

BREAKING CHANGE:
`signMessages` and `strictSigning` pubsub configuration options replaced
with a `globalSignaturePolicy` option

### DIFF
--- a/src/pubsub/index.js
+++ b/src/pubsub/index.js
@@ -540,6 +540,7 @@ class PubsubBaseProtocol extends EventEmitter {
           throw errcode(new Error('Invalid message signature'), codes.ERR_INVALID_SIGNATURE)
         }
         break
+      default:
     }
     for (const topic of message.topicIDs) {
       const validatorFn = this.topicValidators.get(topic)
@@ -566,6 +567,7 @@ class PubsubBaseProtocol extends EventEmitter {
         return signMessage(this.peerId, utils.normalizeOutRpcMessage(message))
       case SignaturePolicy.StrictNoSign:
         return utils.normalizeOutRpcMessage(message)
+      default:
     }
   }
 
--- a/src/pubsub/index.js
+++ b/src/pubsub/index.js
@@ -13,6 +13,7 @@ const { codes } = require('./errors')
  */
 const message = require('./message')
 const PeerStreams = require('./peer-streams')
+const { SignaturePolicy } = require('./signature-policy')
 const utils = require('./utils')
 
 const {
@@ -44,8 +45,7 @@ class PubsubBaseProtocol extends EventEmitter {
    * @param {String} props.debugName log namespace
    * @param {Array<string>|string} props.multicodecs protocol identificers to connect
    * @param {Libp2p} props.libp2p
-   * @param {boolean} [props.signMessages = true] if messages should be signed
-   * @param {boolean} [props.strictSigning = true] if message signing should be required
+   * @param {SignaturePolicy} [props.globalSignaturePolicy = SignaturePolicy.StrictSign] defines how signatures should be handled
    * @param {boolean} [props.canRelayMessage = false] if can relay messages not subscribed
    * @param {boolean} [props.emitSelf = false] if publish should emit to self, if subscribed
    * @abstract
@@ -54,8 +54,7 @@ class PubsubBaseProtocol extends EventEmitter {
     debugName,
     multicodecs,
     libp2p,
-    signMessages = true,
-    strictSigning = true,
+    globalSignaturePolicy = SignaturePolicy.StrictSign,
     canRelayMessage = false,
     emitSelf = false
   }) {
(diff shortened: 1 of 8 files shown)


wemeetagain:
@vasco-santos should be ready for a review

One thing that might be nice to add here, that isn't yet included:
- Using strict no-sign right now requires explicitly passing in a different `msgId` function. It may be nice to provide a default in that case, something with an implementation like `hash(msg.data)`

Do you have a preference for which library/hash could/should be used in this case?

src/pubsub/index.js
```diff
@@ -109,14 +108,17 @@ class PubsubBaseProtocol extends EventEmitter {
      */
     this.peers = new Map()
 
-    // Message signing
-    this.signMessages = signMessages
+    // validate signature policy
+    if (!SignaturePolicy[globalSignaturePolicy]) {
+      throw new Error('Invalid global signature policy')
```
vasco-santos, line 113: Can we also have a code for this?

src/pubsub/index.js
```diff
@@ -511,16 +513,35 @@ class PubsubBaseProtocol extends EventEmitter {
...
+        if (message.key) {
+          throw errcode(new Error('StrictNoSigning: key should not be present'), codes.ERR_UNEXPECTED_KEY)
+        }
+        if (message.seqno) {
+          throw errcode(new Error('StrictNoSigning: seqno should not be present'), codes.ERR_UNEXPECTED_SEQNO)
+        }
+        break
+      case SignaturePolicy.StrictSign:
+        if (!message.signature) {
+          throw errcode(new Error('Signing required and no signature was present'), codes.ERR_MISSING_SIGNATURE)
```
vasco-santos, line 534: Should we prefix the error messages with the policy name, like the `StrictNoSigning`? This would make things more consistent

src/pubsub/index.js
```diff
@@ -511,16 +513,35 @@ class PubsubBaseProtocol extends EventEmitter {
...
+        }
+        if (!message.seqno) {
+          throw errcode(new Error('Signing required and no seqno was present'), codes.ERR_MISSING_SEQNO)
+        }
+        if (!(await verifySignature(message))) {
+          throw errcode(new Error('Invalid message signature'), codes.ERR_INVALID_SIGNATURE)
+        }
+        break
+      default:
```
vasco-santos, line 543: Should we throw here? We will not get here per the validation in the constructor. But, once new policies are added this can get to here if this is not properly updated

src/pubsub/index.js
```diff
@@ -586,13 +611,11 @@ class PubsubBaseProtocol extends EventEmitter {
...
     let msgObject = {
       receivedFrom: from,
-      from: from,
       data: message,
-      seqno: utils.randomSeqno(),
       topicIDs: [topic]
     }
 
-    // ensure that any operations performed on the message will include the signature
+    // ensure that any operations performed on the message will include the signature, if it should exist
```
vasco-santos, line 618:
```suggestion
    // ensure that the message follows the signature policy
```
Seems cleaner this way

src/pubsub/index.js
```diff
@@ -538,11 +559,15 @@ class PubsubBaseProtocol extends EventEmitter {
...
-    } else {
-      return message
+    const signaturePolicy = this.globalSignaturePolicy
+    switch (signaturePolicy) {
+      case SignaturePolicy.StrictSign:
+        message.from = this.peerId.toB58String()
+        message.seqno = utils.randomSeqno()
+        return signMessage(this.peerId, utils.normalizeOutRpcMessage(message))
+      case SignaturePolicy.StrictNoSign:
+        return utils.normalizeOutRpcMessage(message)
```
vasco-santos, line 569: Is this correct? Previously, we did not normalize the message if no sign.
Moreover, [normalizeOutRpcMessage](https://github.com/libp2p/js-libp2p-interfaces/blob/v0.6.0/src/pubsub/utils.js#L97) will create the from property from the previous from property (might throw?)

vasco-santos (changes requested):
Overall this looks good. Just needs some minor changes and docs/examples update.

Can you update https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/pubsub#example to use the policy? Might be worth documenting the constructor params in the API docs as well.
As the end goal is to support topic based policies, I think we should have tests for it, as well as an example in the README for the applications

vasco-santos:
> Using strict no-sign right now requires explicitly passing in a different msgId function. It may be nice to provide a default in that case, something with an implementation like hash(msg.data)

This is not included in go, right? We should leverage [multiformats/js-multihashing-async](https://github.com/multiformats/js-multihashing-async)

wemeetagain:
> This is not included in go, right? We should leverage multiformats/js-multihashing-async

Yea i don't think the go impl includes default msg id functions.

Is there a reason why we need the async version? Somehow it seems that a message id would be something that is always synchronously retrieved.

vasco-santos:
> Is there a reason why we need the async version? Somehow it seems that a message id would be something that is always synchronously retrieved.

Good point, we should use the base module then: https://github.com/multiformats/js-multihash

vasco-santos:
@wemeetagain Can you get the lint job fixed?

vasco-santos (approved):
LGTM

wemeetagain:
bump

jacobheun (approved):
LGTM 🚢

PR #67, opened 2020-10-15
feat: support for stream muxer closeRead and closeWrite

jacobheun:
This adds tests for closeRead and closeWrite on the muxer streams.

Also:
- Removed `libp2p-tcp` in favor of `it-pair` to avoid circular deps
- `Connection.close()` will now close its internal streams to avoid lingering streams.

BREAKING CHANGE: This adds closeWrite and closeRead checks in the tests, which will cause test failures for muxers that dont implement those.


achingbrain:
Superseded by #90

PR #69, opened 2020-11-03
pubsub: fix typescript types

wemeetagain:
Unfortunately, #66, as merged, had several errors with the typescript types:
- the globalSignaturePolicy type was set as `SignaturePolicy`, rather than `string`
  - the idea is that `SignaturePolicy` is the "type" of the enum, which is how this would be described if this was written in typescript. Unfortunately, this type was not exported so there was a type mismatch.
- the checked-in typescript declaration files were not updated

This PR rectifies both issues


vasco-santos (approved):
LGTM!

PR #71, opened 2020-11-25
chore: add duplex iterable type to connection

vasco-santos:
This PR adds the duplex iterable type to connection per https://github.com/libp2p/js-libp2p/pull/802#discussion_r524290682

This already existed in the pubsub peer streams and it was moved into the conenction and linked from there

cc @wemeetagain (could not ask you direct review)

### DIFF
--- a/src/connection/connection.d.ts
+++ b/src/connection/connection.d.ts
@@ -1,5 +1,24 @@
 declare const _exports: typeof Connection;
 export = _exports;
+export type Sink = (source: Uint8Array) => Promise<Uint8Array>;
+export type DuplexIterableStream = {
+    sink: Sink;
+    source: () => AsyncIterator<Uint8Array, any, undefined>;
+};
+export type PeerId = import("peer-id");
+export type Multiaddr = import("multiaddr");
+/**
+ * @callback Sink
+ * @param {Uint8Array} source
+ * @returns {Promise<Uint8Array>}
+ *
+ * @typedef {object} DuplexIterableStream
+ * @property {Sink} sink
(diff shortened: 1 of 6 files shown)


src/connection/connection.d.ts
```diff
@@ -1,5 +1,24 @@
 declare const _exports: typeof Connection;
 export = _exports;
+export type Sink = (source: Uint8Array) => Promise<Uint8Array>;
+export type DuplexIterableStream = {
+    sink: Sink;
+    source: () => AsyncIterator<Uint8Array, any, undefined>;
+};
```
jacobheun, line 7: The type here should really just be a `Stream`, and it should match what's coming from mplex https://github.com/libp2p/js-libp2p-mplex/blob/v0.10.1/src/stream.js#L59

vasco-santos: Should we add the Stream in a type declaration file in the multiplexer interface?

src/connection/connection.js
```diff
@@ -1,56 +1,23 @@
...
-  }
-}
+/**
+ * @callback Sink
+ * @param {Uint8Array} source
+ * @returns {Promise<Uint8Array>}
+ *
+ * @typedef {object} DuplexIterableStream
+ * @property {Sink} sink
+ * @property {() AsyncIterator<Uint8Array>} source
```
Gozala, line 16: Is this `AsyncIterator` or is it meant to be `AsyncIterable` ?

vasco-santos:
Closed in favour of #74

PR #74, opened 2020-12-01
feat: add types

vasco-santos:
This PR adds typedef improvements for the already available modules, as well as typescript files for the interfaces needed in libp2p core per https://github.com/libp2p/js-libp2p/pull/802

- Aegir updated to latest version to support types generation
  - resulting linting issues fixed
- Types for interfaces: Transport, Muxer and Crypto
  - should probably look into uniforming the language on the types and Readme
- Switched from travis to Github actions

On top of @Gozala 's #73 PR

Needs:
- [x] type declaration files removed (currently in place to be used by libp2p PR)

BREAKING CHANGE: Some types from connection, pubsub and topology are now fixed. Record is not a class anymore, but a typescript interface that should be implemented


src/transport/types.ts
```diff
@@ -0,0 +1,120 @@
+import events from 'events'
+
+/**
+ * A libp2p transport is understood as something that offers a dial and listen interface to establish connections.
+ */
+export interface Interface {
```
Gozala, line 6: More concrete name would be much better here.

src/transport/types.ts
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Upgrades an inbound connection on transport listener.
+   *
+   * @param {MultiaddrConnection} maConn
+   * @returns {Promise<Connection>}
+   */
+  upgradeInbound(maConn: MultiaddrConnection): Promise<Connection>;
+}
+
+export declare class Transport implements Interface {
```
Gozala, line 71: I think this should be an interface as well not a class.

tsconfig.json
```diff
@@ -6,6 +6,20 @@
     // Tells TypeScript to read JS files, as
```
Gozala, line 6: we should replace this with `aegir ts --preset config > tsconfig.json` instead.

Gozala (changes requested):
@vasco-santos I did a pass here, but I think I have a more general feedback that is not pinned to any specific line:

- `aegir bulid` will generate types in `dist` so no `.d.ts` files are necessary (I did pushed a commit here that removes them, but feel free to revert it if you disagree).
- types in `dist` need to be mapped in package.json (I have created #75  that does it).
- type check / build is failing with this tree (I have created #75 that addresses remaining issues)
- I think we need to distinguish between interfaces (`interface { ...}`) and classes that implement those. And users of those implementations should code against interfaces not implementation classes. That enables anyone to do a different implementation of the interface, ensure with type checker compatibility and use it as drop-in replacement.
- I think `declare class`-es must go, if they are necessary they should be interfaces.
- There are some incompatibilities regarding `MuxedStream` and `DuplexIterableStream`s (I had to suppress some mismatches in #75). I think we should resolve those. I don't have a complete enough understanding in this area to provide specific suggestions, but would love to have session so I can better understand it and also help resolve this.
- There are bunch of places where instance fields are `null|T` but in code used as `T` (I have suppressed those with `@ts-ignore` in #75, but I think we need a followup pull to resolve them)

src/pubsub/peer-streams.js
```diff
@@ -123,11 +133,13 @@ class PeerStreams extends EventEmitter {
     // - transformed with length-prefix transform
     this._inboundAbortController = new AbortController()
     this._rawInboundStream = stream
+    // @ts-ignore - abortable returns AsyncIterable and not a DuplexIterableStream
```
hugomrdias, line 136: i think the type `DuplexIterableStream` should be improved because its just an object and not a iterable or a stream :)

src/pubsub/peer-streams.js
```diff
@@ -43,35 +46,41 @@ class PeerStreams extends EventEmitter {
...
      * @private
-     * @type {typeof AbortController}
+     * @type {null|AbortController}
      */
     this._inboundAbortController = null
     /**
      * Write stream -- its preferable to use the write method
-     * @type {import('it-pushable').Pushable<Uint8Array>>}
+     *
+     * @type {null|import('it-pushable').Pushable<Uint8Array>}
```
hugomrdias, line 77: don't import inline, this makes some tools confused and the .d.ts generated more complex

src/pubsub/message/sign.js
```diff
@@ -31,14 +31,16 @@ async function signMessage (peerId, message) {
...
  * @param {InMessage} message
- * @returns {Promise<Boolean>}
+ * @returns {Promise<boolean>}
  */
 async function verifySignature (message) {
   // Get message sans the signature
   const baseMessage = { ...message }
   delete baseMessage.signature
   delete baseMessage.key
+  // @ts-ignore - from is optional
```
hugomrdias, line 43: its safer to either make it required or protect with and if clause

src/pubsub/message/sign.js
```diff
@@ -49,6 +51,7 @@ async function verifySignature (message) {
   const pubKey = await messagePublicKey(message)
 
   // verify the base message
+  // @ts-ignore - may not have signature
```
hugomrdias, line 54: its safer to either make it required or protect with and if clause

src/pubsub/message/sign.js
```diff
@@ -61,6 +64,7 @@ async function verifySignature (message) {
  */
 async function messagePublicKey (message) {
   // should be available in the from property of the message (peer id)
+  // @ts-ignore - from is optional
```
hugomrdias, line 67: its safer to either make it required or protect with and if clause

src/pubsub/index.js
```diff
@@ -696,6 +730,16 @@ class PubsubBaseProtocol extends EventEmitter {
...
+ * @typedef {import('../connection/connection')} Connection
+ * @typedef {import('./message').RPC} RPC
+ * @typedef {import('./message').SubOpts} RPCSubOpts
+ * @typedef {import('./message').Message} RPCMessage
+ * @typedef {import('./signature-policy').SignaturePolicyType} SignaturePolicyType
+ */
+
 module.exports = PubsubBaseProtocol
 module.exports.message = message
 module.exports.utils = utils
```
hugomrdias, line 745: its better to always use named exports especially in this case where default and named are mixed!

src/pubsub/index.js
```diff
@@ -317,16 +331,18 @@ class PubsubBaseProtocol extends EventEmitter {
...
+   * @param {DuplexIterableStream} stream - inbound stream
+   * @param {PeerStreams} peerStreams - PubSub peer
    * @returns {Promise<void>}
    */
   async _processMessages (idB58Str, stream, peerStreams) {
     try {
       await pipe(
         stream,
         async (source) => {
+          // @ts-ignore - DuplexIterableStream isn't defined as iterable
```
hugomrdias, line 345: DuplexIterableStream needs to be extracted to types.ts file since its used in multiple places and improved to be an iterable

src/pubsub/peer-streams.js
```diff
@@ -43,35 +46,41 @@ class PeerStreams extends EventEmitter {
...
      */
     this._rawOutboundStream = null
     /**
      * The raw inbound stream, as retrieved from the callback from libp2p.handle
+     *
      * @private
-     * @type {DuplexIterableStream}
+     * @type {null|DuplexIterableStream}
      */
     this._rawInboundStream = null
```
hugomrdias, line 66:
```suggestion
     * @type {DuplexIterableStream}
     */
    this._rawInboundStream
```

dont init to null or else you will need to protect everywhere this is used, just https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#classes

src/pubsub/index.js
```diff
@@ -205,26 +212,29 @@ class PubsubBaseProtocol extends EventEmitter {
...
-   * @param {Connection} props.connection connection
+   * @param {Connection} props.connection - connection
    */
   _onIncomingStream ({ protocol, stream, connection }) {
     const peerId = connection.remotePeer
     const idB58Str = peerId.toB58String()
     const peer = this._addPeer(peerId, protocol)
     peer.attachInboundStream(stream)
 
+    // @ts-ignore - peer.inboundStream maybe null
```
hugomrdias, line 228: see comment in peer-streams

src/crypto/types.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export type SecureOutbound = {
+  conn: MultiaddrConnection;
+  remoteEarlyData: Buffer;
+  remotePeer: PeerId;
+}
+
+type PeerId = import('peer-id');
+type MultiaddrConnection = import('../transport/types').MultiaddrConnection
```
hugomrdias, line 40: if not exporting just import at the top

src/connection/connection.js
```diff
@@ -198,37 +192,96 @@ class Connection {
...
+ * @property {Timeline} timeline - connection relevant events timestamp.
+ * @property {string} [multiplexer] - connection multiplexing identifier.
+ * @property {string} [encryption] - connection encryption method identifier.
+ *
+ * @typedef {Object} Timeline
+ * @property {number} open - connection opening timestamp.
+ * @property {number} [upgraded] - connection upgraded timestamp.
+ * @property {number} [close]
+ *
+ * @typedef {import('./status').Status} Status
```
hugomrdias, line 238: can we put typedef all in the same place either top or bottom ?

hugomrdias (changes requested):

vasco-santos:
Thanks for the reviews. On @Gozala feedback:

> aegir bulid will generate types in dist so no .d.ts files are necessary (I did pushed a commit here that removes them, but feel free to revert it if you disagree).
> types in dist need to be mapped in package.json (I have created #75 that does it).

Yes, as mentioned on the PR desc, I kept them to test it with https://github.com/libp2p/js-libp2p/pull/802 
Unfortunately this will be a limitation for testing stuff across multiple PR dependencies.

> I think we need to distinguish between interfaces (interface { ...}) and classes that implement those. And users of those implementations should code against interfaces not implementation classes. That enables anyone to do a different implementation of the interface, ensure with type checker compatibility and use it as drop-in replacement.
I think declare class-es must go, if they are necessary they should be interfaces.

I added them to go around the type checker error of the type not being constructable. Anyway, this did not fix the issue, so I will remove them.

> There are some incompatibilities regarding MuxedStream and DuplexIterableStreams (I had to suppress some mismatches in #75). I think we should resolve those. I don't have a complete enough understanding in this area to provide specific suggestions, but would love to have session so I can better understand it and also help resolve this.

Yeah, I think they should all be MuxedStream, if you are talking about Pubsub. I will change them

> There are bunch of places where instance fields are null|T but in code used as T (I have suppressed those with @ts-ignore in #75, but I think we need a followup pull to resolve them)

I will have a look on these

src/record/index.js
```diff
@@ -7,25 +7,31 @@ const errcode = require('err-code')
  */
 class Record {
   /**
-   * @constructor
-   * @param {String} domain signature domain
-   * @param {Uint8Array} codec identifier of the type of record
+   * @class
```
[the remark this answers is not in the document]
vasco-santos: yeah, it would be good to change it and to auto fix opposite

src/transport/types.ts
vasco-santos: removing the class, I will be calling Transport

src/transport/types.ts
vasco-santos: > if the purpose here is to type the constructor

It is and it would work for the transports. However js-libp2p uses creates its configured transports as `new Transport()` and not `new TCP()`, so the type there should be `Transport`. I think we can go with the base class in a follow up PR, as this will need all the transports to also be changed. 
We are targeting a first iteration with just support in the core.

I will use the interface for now

src/transport/types.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+export interface Interface {
+  /**
+   * Dial a given multiaddr.
+   *
+   * @param {Multiaddr} ma
+   * @param {any} [options]
+   * @returns {Promise<Connection>}
+   */
+  dial(ma: Multiaddr, options?: any): Promise<Connection>;
```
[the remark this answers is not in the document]
vasco-santos: Very specific for each transport. The signal option for supporting aborts is the only option specified by the interface, then each transport might have completely different options.

It should be signal and any others, what is the best way to type this?

src/topology/multicodec-topology.js
```diff
@@ -46,7 +38,25 @@ class MulticodecTopology extends Topology {
...
+
+  get [multicodecTopologySymbol] () {
+    return true
+  }
+
+  /**
+   * Checks if the given value is a `MulticodecTopology` instance.
+   *
+   * @param {any} other
+   * @returns {other is MulticodecTopology}
```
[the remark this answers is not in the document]
vasco-santos: not sure, @Gozala ?

src/stream-muxer/tests/spawner.js
```diff
@@ -78,5 +84,7 @@ function marker (n) {
       }
     }
   })
+
+  // @ts-ignore - TS can't see that assignement occured
```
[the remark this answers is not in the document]
vasco-santos: yap, I think we can keep the ts ignore for now as these are stream muxer tests. Once we update mplex with types, we can also have a look back in these tests

src/record/index.js
```diff
@@ -7,25 +7,31 @@ const errcode = require('err-code')
...
   /**
    * Marshal a record to be used in an envelope.
+   *
+   * @returns {Uint8Array}
    */
   marshal () {
     throw errcode(new Error('marshal must be implemented by the subclass'), 'ERR_NOT_IMPLEMENTED')
   }
 
+  // eslint-disable-next-line
```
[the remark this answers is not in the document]
vasco-santos: The issue here is the return below. We use this in libp2p generically in the https://github.com/libp2p/js-libp2p/blob/master/src/record/envelope/index.js Envelope.seal(). If we do not set here the types returned it would be void and cause issues there

src/record/index.js
```diff
@@ -7,25 +7,31 @@ const errcode = require('err-code')
...
   /**
    * Verifies if the other provided Record is identical to this one.
+   *
    * @param {Record} other
+   * @returns {boolean}
```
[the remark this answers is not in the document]
vasco-santos: The issue here is the return below. We use this in libp2p generically for any type of libp2p record implemented. If we do not set here the types returned it would be void and cause issues there

src/record/index.js
```diff
@@ -7,25 +7,31 @@ const errcode = require('err-code')
  */
 class Record {
```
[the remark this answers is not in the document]
vasco-santos: This is extended in https://github.com/libp2p/js-libp2p/tree/master/src/record/peer-record . We can also have a types definition, but it seems redundant. What do you think?

src/pubsub/message/sign.js
vasco-santos: Per the [spec](https://github.com/libp2p/specs/tree/master/pubsub#the-message), it must be optional. So I will be adding an if clause

src/pubsub/message/sign.js
vasco-santos: Cannot be required by multiple signing policies in the pubsub spec. Will add an if

src/pubsub/message/sign.js
vasco-santos: It is optional in fact, but when it reaches this function, it will exist anyway. So, it seems better to do the ts ignore than to protect every usage, which will also lead to a need of new return type

src/pubsub/peer-streams.js
vasco-santos: This is a good practise that we can use. In this case it is more complex because multiple inbound streams will be attached over time to the peer.
If we have it in the constructor when we create the peer, we need to take into account only follow up attachments which makes this logic more complex. IMHO we should not add complexity to the code to favour internal types

src/connection/connection.js
vasco-santos: yes, moving them

src/pubsub/peer-streams.js
hugomrdias: this is basically just trading null for undefined does it make a big difference ?

src/pubsub/peer-streams.js
vasco-santos: It was changed to a MuxedStream as it is what it should be

src/pubsub/index.js
vasco-santos: It should really be a muxedStream. Updated it

src/record/index.js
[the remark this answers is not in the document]
Gozala: > @constructor can be useful https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#constructor

From what I can tell it only makes diff when you use non-class syntax to define classes. I’m pretty sure TS forbids calling without `new` when defined with class syntax.

So I’d suggest removing those

src/transport/types.ts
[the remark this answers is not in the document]
Gozala: You should make it generic 

```ts
interface Transport<DialOptions> {
  dial(address: Multiaddress, options: DialOptions): Promise<Connection>
}
```

src/pubsub/peer-streams.js
vasco-santos: Do you mean changing this to to equal undefined? It is a js file, I cannot just do the suggested change

src/transport/types.ts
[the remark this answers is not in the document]
vasco-santos: I cannot add it with the following:

```js
/**
 * A libp2p transport is understood as something that offers a dial and listen interface to establish connections.
 */
export interface Transport {
  new (upgrader: Upgrader, ...others: any): Transport; // eslint-disable-line
  /**
   * Dial a given multiaddr.
   */
  dial(ma: Multiaddr, options?: any): Promise<Connection>;
  /**
   * Create transport listeners.
   */
  createListener(options: any, handler: (Connection) => void): Listener;
  /**
   * Takes a list of `Multiaddr`s and returns only valid addresses for the transport
   */
  filter(multiaddrs: Multiaddr[]): Multiaddr[];
}
```

The constructor would need the DialOptions. It is really part of the dial function, not constructor related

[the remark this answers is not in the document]
vasco-santos: New fails type check with: `Generic type 'Transport<DialOpions>' requires 1 type argument(s).ts(2314)`

src/pubsub/index.js
```diff
@@ -696,7 +728,19 @@ class PubsubBaseProtocol extends EventEmitter {
   }
 }
 
+/**
+ * @typedef {any} Libp2p
+ * @typedef {object} MuxedStream
+ * @type import('../stream-muxer/types').MuxedStream
```
hugomrdias, line 734:
```suggestion
 * @typedef {import('../strea-muxer/types').MuxedStream}MuxedStream
```

src/pubsub/index.js
```diff
@@ -328,6 +341,7 @@ class PubsubBaseProtocol extends EventEmitter {
         stream,
         async (source) => {
           for await (const data of source) {
+            // @ts-ignore data slice from BufferList
```
hugomrdias, line 344: add bl types to MuxedStream.source 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bl/index.d.ts

src/pubsub/message/sign.js
```diff
@@ -61,13 +65,14 @@ async function verifySignature (message) {
  */
 async function messagePublicKey (message) {
   // should be available in the from property of the message (peer id)
+  // @ts-ignore - from type changed
   const from = PeerId.createFromCID(message.from)
```
hugomrdias, line 69: protect here ?

src/pubsub/peer-streams.js
```diff
@@ -17,20 +20,22 @@ log.error = debug('libp2p-pubsub:peer-streams:error')
...
- * @property {Sink} sink
- * @property {() AsyncIterator<Uint8Array>} source
+ * @typedef {object} MuxedStream
+ * @type import('../stream-muxer/types').MuxedStream
  *
  * @typedef PeerId
  * @type import('peer-id')
+ *
+ * @typedef PushableStream
+ * @type import('it-pushable').Pushable<Uint8Array>
```
hugomrdias, line 30: can you fix this syntax 
`@typedef {import('../stream-muxer/types').MuxedStream} MuxedStream`

src/stream-muxer/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  onStreamEnd: (stream: MuxedStream) => void;
+  maxMsgSize?: number;
+}
+
+export type MuxedTimeline = {
+  open: number;
+  close?: number;
+}
+
+export type MuxedStream = {
```
hugomrdias, line 36: this should extend AsyncIterable

src/transport/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  sink: Sink;
+  source: () => AsyncIterable<Uint8Array>;
+  close: (err?: Error) => Promise<void>;
+  conn: any;
+  remoteAddr: Multiaddr;
+  localAddr?: Multiaddr;
+  timeline: MultiaddrConnectionTimeline;
+}
+
+export type Sink = (source: Uint8Array) => Promise<Uint8Array>;
```
hugomrdias, line 69: this is already in stream-muxer types

hugomrdias (changes requested):

src/pubsub/peer-streams.js
```diff
@@ -107,22 +109,23 @@ class PeerStreams extends EventEmitter {
...
    * @returns {void}
    */
   attachInboundStream (stream) {
     // Create and attach a new inbound stream
     // The inbound stream is:
     // - abortable, set to only return on abort, rather than throw
     // - transformed with length-prefix transform
-    this._inboundAbortController = new AbortController()
     this._rawInboundStream = stream
+    // @ts-ignore - abortable returns AsyncIterable and not a MuxedStream
```
hugomrdias, line 128: this can be removed now that MuxedStream extends AsyncIterable

hugomrdias (changes requested):
i think this is the last bit

src/pubsub/peer-streams.js
vasco-santos: In fact, the abort receives the MuxedStream and returns a AsyncIterable<any>. So, as `this.inboundStream` is always wrapped with the abort I changed it from `MuxedStream` to `AsyncIterable<Uint8Array>`

hugomrdias (approved):

hugomrdias:
Great work!! 💪🏼

src/transport/types.ts
Gozala: I think it is important to distinguish between type (interface) of the value and how do you create one. In most places you'd require value of certain type to be provided regardless how it was created & in much fewer places (if any) you would require a factory for creating these values. The fact that factory happens to be the class / constructor for the type should really be irrelevant.

That is to suggest that we should separate those concerns and decouple them from each other even if in practice both are implemented via same class definition.

src/transport/types.ts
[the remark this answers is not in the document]
Gozala: I'm not sure where you seeing it so it's hard for me to provide any meaningful feedback. Maybe let's try to resolve this in a higher bandwidth channel ?

src/transport/types.ts
vasco-santos: @hugomrdias found a way of adding the constructor in the interface. So this seems good now just with the interface

src/transport/types.ts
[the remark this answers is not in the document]
Gozala: So you could have a class like this:

```ts
/**
 * @implements {Transport<{ foo: boolean }>}
 */
class FooTransport {
   /**
    * @param {Multiaddr} address
    */
  constructor(address) {
  }
  dial(ma: Multiadddr, options?: DialOptions) {
  }
}
```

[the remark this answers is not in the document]
Gozala: Based on the discussion in other channel

> TCP will eventually accept some given options for tcp.dial(multiaddr, options), which websocket.dial(multiaddr, otherOptions).
What the interface states is that any transport on .dial must provide signal for Abort controller as a property, while the others are something else
> we want something like: { signal, …other } to be the DialOptions

That could be achieved as follows:

```ts
interface Transport<DialOptions extends {signal?: AbortSignal}> {
  dial(address: Multiaddress, options: DialOptions): Promise<Connection>
}
```

Than in the implementation you can have something like this:

```ts
type TCPDialOptions = {
  signal?: AbortSignal
  tcpOptionThingy: ...
}

class TCPTransport implements Transport<TCPDialOptions> {
  constructor(options: ThingUnrelatedToDialOptions) {
     // ...
  }
  dial(address:Multiaddress, options: TCPDialOptions) {
    // ...
  }
}
```

Note: However that you can't make `options` param on dial optional in some instances and non optional in other. However if they are optional everywhere that you could just make it optional as `dail(address:Multiaddress, options?:DialOptions)`

src/pubsub/peer-streams.js
Gozala: I think making those types non optional as in `Stream` over `Stream|null` (or `Stream|undefined`)  is a bad idea is it will prevent type checker from identifying real errors. E.g. some method that assumes `_rawInboundStream` might be called before it is initialized and type checker will not remind you to check that value is there.

I am assuming however that root of the problem [boils down to this example](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAKCm8BOBlALk+YC2MDeAsAFDTQD2ArmgEaUB2AJupjgFzR0UgjQA+0AdSQBLNGGoh4zLNmKlgZOhAwVgaMkgAUASmiESpaGgAWwiADpKNekwwzoAXg5cQc6AF83Ac3hpoZoVFxSWgdPTdDTDQKJDpoACEyMkkwOk0TM0sqWgpGaRxtN08DAHcRNHhQhjAxVgBVYTo0AA4AQSQkMABPXX1DfwAzUIBCDIsA8uD4XojDEyQyEo54JYBRDo1NAHIAOTJoMqCJSoU6Ong1YUUtwoNSYtmxrOtc2xZscx80QIqtbXNDhVNNUxLd+gB6cGGAB60OgAHlqAArC5+MzQAAOZCgwgkXWgW043C2RWI7iAA) (simplified so it's easy to throw into ts playground):

```ts
class PeerStreams {
  outboundStream: null | WritableStream
  constructor() {
    this.outboundStream = null
  }
  get isWritable () {
     return Boolean(this.outboundStream)
  }
  write (data:Uint8Array) {
    if (!this.isWritable) {
      throw new Error('No writable connection')
    }

    this.outboundStream.getWriter().write(data)
    //   ^^ Object is possibly 'null'
  }
}
```

Where it is clear that `this.outboundStream` isn't null because of `!this.isWritable` check above. Unfortunately TS can't (yet) derive that because it does not know about relation here. It could however be told to [recognize that as follows](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAKCm8BOBlALk+YC2MDeAsAFDTQD2ArmgEaUB2AJupjgFzR0UgjQA+0AdSQBLNGGoh4zLNmKlgZOhAwVgaMkgAUASmiESpaGgAWwiADpKNekwwzoAXg5cQc6AF83Ac3hpoZoVFxSWgdPTdDTDQKJDp-CECxCXhNEzNtN08DAHcRNHhQhjAxVgBVYTo0AA4AQSQkMABPXX1DfwAzUIBCALzglLSIbRaIwxMkMmyOeCmAUXqNTQByADkyaFyg5OgFOjp4NWFFJYyDUizRwcsqWgpGaRxzHzRE5B1zTfzNIrFT8+ILkRdsp4ol+o5QsoWNhWABZJrUeBg5LadhQ+xmQR9bYOAB8bgAQmQyJIwHRNOjHlZbvc7DhTsQKvkkO0wMACvDGojkSFWuQbjYHjCsVtJEK+M5uADiIzKshWeyRUleW5qYK6cKeVINdKiEA):

```ts
class PeerStreams {
  outboundStream: null | WritableStream
  constructor() {
    this.outboundStream = null
  }
  get isWritable () {
     return isWritable(this)
  }
  write (data:Uint8Array) {
    if (!isWritable(this)) {
      throw new Error('No writable connection')
    }

    this.outboundStream.getWriter().write(data)
  }
}

const isWritable = (stream:MaybeWritable): stream is Writable =>
  Boolean(stream.outboundStream)

interface MaybeWritable {
  outboundStream: WritableStream | null
}

interface Writable {
  outboundStream: WritableStream
}
```

P.S: Unfortunately you `: stream is Writable` types annotations don't seem to be (yet) supported on getters, but I do hope that helper function is not a large price to pay for the added benefit, especially since they could probably be reused elsewhere.

src/connection/connection.js
Gozala: I would also just move those defs into .ts file

Gozala:
> Yes, as mentioned on the PR desc, I kept them to test it with [libp2p/js-libp2p#802](https://github.com/libp2p/js-libp2p/pull/802)
> Unfortunately this will be a limitation for testing stuff across multiple PR dependencies.

I think that could (and should) be address by adding a `prepare` npm script that generates types. I think we even have some packages that do that.

Gozala:
> I added them to go around the type checker error of the type not being constructable. Anyway, this did not fix the issue, so I will remove them.

I wrote about it at length in the here https://github.com/libp2p/js-libp2p-interfaces/pull/74/files/67a5f518050e847799272a788d7182bd73a9976f#r534350970

I do still maintain the position that it is very useful to separate concerns and corresponding interfaces for a type and it's constructor, because again user of the type is different from the user of the constructor and separation of those concerns loosens requirements e.g. I could write an implementation with incompatible constructor and that is no reason for user of the type to reject it.

src/stream-muxer/types.ts
Gozala:
```suggestion
export interface MuxedStream extends AsyncIterable<Uint8Array> {
```

src/stream-muxer/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  close: () => void;
+  abort: () => void;
+  reset: () => void;
+  sink: Sink;
+  source: () => AsyncIterable<Uint8Array>;
+  timeline: MuxedTimeline;
+  id: string;
+  [Symbol.asyncIterator](): AsyncIterator<Uint8Array>;
```
Gozala: With the above suggestion applied this is no longer neeeded
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,53 @@
+name: ci
```
hugomrdias, line 1: fixed some bugs in the gh actions please refresh this file with https://github.com/ipfs/aegir/wiki/Github-Actions-Setup

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,27 @@
+on:
```
hugomrdias, line 1: this file can be removed now that we have main.yml

hugomrdias (changes requested):

src/transport/types.ts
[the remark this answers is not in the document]
vasco-santos: Ok, this seems to work! Added it, thank you 👌

src/stream-muxer/types.ts
vasco-santos: I think this was addressed in my latest commits, also added BufferList

src/stream-muxer/types.ts
vasco-santos: I think this was addressed in my latest commits, also added BufferList

src/record/index.js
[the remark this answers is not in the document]
vasco-santos: Yeah, I would just add that support in the linter so that we remove it from everywhere automatically

vasco-santos:
> I wrote about it at length in the here https://github.com/libp2p/js-libp2p-interfaces/pull/74/files/67a5f518050e847799272a788d7182bd73a9976f#r534350970

> I do still maintain the position that it is very useful to separate concerns and corresponding interfaces for a type and it's constructor, because again user of the type is different from the user of the constructor and separation of those concerns loosens requirements e.g. I could write an implementation with incompatible constructor and that is no reason for user of the type to reject it.

Github is broken already, I cannot open the link you reference. Anyway, I think that your point was:

> I think it is important to distinguish between type (interface) of the value and how do you create one. In most places you'd require value of certain type to be provided regardless how it was created & in much fewer places (if any) you would require a factory for creating these values. The fact that factory happens to be the class / constructor for the type should really be irrelevant.

> That is to suggest that we should separate those concerns and decouple them from each other even if in practice both are implemented via same class definition.

So, my suggestion here is that we punt on this until we get the types into the transports codebase. By that time, we can really see the advantages/limitations and decide if we should add here a second interface for the constructor. What do you thing?

FWIW, adding types to other modules will happen over time, but it not a priority now. Users will only interact with libp2p core API unless they are creating new Transports, Stream Muxers, ... We just really need to guarantee in libp2p that the interface is compliant and it is constructable

Gozala:
> So, my suggestion here is that we punt on this until we get the types into the transports codebase. By that time, we can really see the advantages/limitations and decide if we should add here a second interface for the constructor. What do you thing?

Sure if you think this is a better way to go.

Gozala:
@hugomrdias @vasco-santos I'm not sure why type check on CI was disabled, but I think it is not a good idea, because it provides feedback when there is a mismatch. It is true that `aegir bulid` will perform type check, but it will not report missmatches in the pull diff, which I find to be a lot more convenient than it is to read through CI logs.

vasco-santos:
> I'm not sure why type check on CI was disabled, but I think it is not a good idea, because it provides feedback when there is a mismatch. It is true that aegir bulid will perform type check, but it will not report missmatches in the pull diff, which I find to be a lot more convenient than it is to read through CI logs.

If the aegir build does not provide that diff I agree that we should add it back. @hugomrdias ?

src/pubsub/index.js
```diff
@@ -205,26 +221,28 @@ class PubsubBaseProtocol extends EventEmitter {
...
+   * @param {Connection} props.connection - connection
    */
   _onIncomingStream ({ protocol, stream, connection }) {
     const peerId = connection.remotePeer
     const idB58Str = peerId.toB58String()
     const peer = this._addPeer(peerId, protocol)
     peer.attachInboundStream(stream)
 
-    this._processMessages(idB58Str, peer.inboundStream, peer)
+    peer.inboundStream && this._processMessages(idB58Str, peer.inboundStream, peer)
```
Gozala, line 237: It might be worth to make `attachInboundStream(stream)` return the value that it's assigning to the `this.inboundStream` that way this will turn into:

```js
const inboundStream = peer.attachInboundStream(stream)
this._processMessage(idB58Str, inboundStream, peer)
```

src/pubsub/peer-streams.js
```diff
@@ -107,21 +109,21 @@ class PeerStreams extends EventEmitter {
       throw new Error('No writable connection to ' + id)
     }
 
+    // @ts-ignore - this.outboundStream could be null
```
Gozala, line 112: Relevant comment here
https://github.com/libp2p/js-libp2p-interfaces/pull/74#discussion_r534383557

src/pubsub/peer-streams.js
```diff
@@ -138,26 +140,26 @@ class PeerStreams extends EventEmitter {
...
    * Attach a raw outbound stream and setup a write stream
    *
-   * @param {Stream} stream
+   * @param {MuxedStream} stream
    * @returns {Promise<void>}
    */
   async attachOutboundStream (stream) {
     // If an outbound stream already exists,
     // gently close it
     const _prevStream = this.outboundStream
```
Gozala, line 149: In that case this line is no longer needed
```suggestion
```

src/record/index.js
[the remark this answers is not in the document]
Gozala: If thing does not need to be a class it's best to define it as interface / type because it could be substituted with other implementation or even just plain object that conforms the interface. Classes should really just be a one possible implementation of it. Current version implies that TS will error anywhere `record instanceof Record` isn't true.

src/topology/multicodec-topology.js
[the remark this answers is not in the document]
Gozala: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

src/transport/types.ts
```diff
@@ -0,0 +1,73 @@
...
+export type DialOptions = {
+  signal?: AbortSignal
+}
+
+/**
+ * A libp2p transport is understood as something that offers a dial and listen interface to establish connections.
+ */
+export interface Transport <DialOptions extends { signal?: AbortSignal }> {
+  new (upgrader: Upgrader, ...others: any): Transport<DialOptions>; // eslint-disable-line
+  prototype: Transport <DialOptions>;
```
Gozala, line 15: Is this really necessary ? If not I'd get rid of it

Gozala (approved):
Add few nits, but looks good to me to go with or without them.

Gozala:
> So, my suggestion here is that we punt on this until we get the types into the transports codebase. By that time, we can really see the advantages/limitations and decide if we should add here a second interface for the constructor. What do you thing?

Seeing the `prototype` field and `new` in the interface definition, I do really think it would be an easier change to do now than deferring it to a future. It's going to be a couple of lines change but avoid having to propagate changes in the future. I'll provide specific comment in the pull.

Either way it's just a suggestion and it's your decision to make.

Gozala:
Ok if I take the current typedefs for transports and try to [implement it like this](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAUwG4IHYwM5wGZQiOAcmTU0IFgAoUSWOAWQFcAbGYAQwBNOoc8DCIFmy48K1dAijZ2AYwRwAwhFSoEstirgBvOAHcowGAgAUndjHYAuAKrB0ADgCCUKOwCeASitIIwTgC+VDDuYAoAyvYA1nAAvHAmGBCMUPJWcHaOLm5ecQB8cAAK-MAYCAA8mTDOrh55ANxUwaEKACqKhQAiHMxxOnAYwADmqOzMAPzpTgBG0DCRI2NwQZQIAB608LLM7BhY7YWtbqgYm3A0zAggpPvHp3PlB91jBdpUAPTvcN8-v98Aev8lDs9kQDkd2CdNoRzqhZNAoOoYMx3OdwJdrugsPZjNI5ApCBCoQ8nj08oQAHRUb7wk4wKCMDTQEyMMBDNycKTpGxsjlSAA0cApwogMAAFlIMOlIbk3pRfoiYClUHBxaVqcsNe8AFTajVwbVwZ69dhwIbAFAqoSsDjcKBU+XfbXvDWcHomEDWBjCW08QUQMCaE6TOCksbeIolMrlZSqJHAFSvfWqsV4PRwACMGpW3x1esdBqUiIsCnpkPudGYpWMaigGAdP2dGtkxeMABlq2gpCYA0GpXAZYKxZDOJcoOkTLS1BoE6grLHp0HPPk4L5-BGOxga1IdMnxWm4AAmbNa3X6w2tdhRBBYU1VrdwCDYOAAAyYNtEUBft9QnDgiuVLAVBRVcxn8Ac7RvMosGwaAU1LO5NnPF1HWwYBWG7a0RDtft32wngAG0AF0Izw30oGI3cC3-BAlSgK0fU-DAT0oFYqHWM4QjCI0egAeUDWcsHiOVvkGRYJimWZYAWUZmCoNjKDzalDScOAq2mMBDzAVVELmc4sEYX9JRgCAID-XYBnwWixXsIYUwsR9sGwSUBzgN0lhHNTOxVHEpBkeRVQgRAt3YaZ7zFOAp3jFR62UlCOL03y8QCokK3gcpjX4vtEDWGtOCwXQxNkkMZjmGSlgCJNHTUdMWV5LguQyerOSgQVhQpUUJTraVUC8dJUs2DK+IEmKGjgT5gqrdAAFo3QwULLmmqaEA1MA8BMriEH63S6CGsYssEhpT3zRseI8s0LTQOAsPIhsnRQ753OYD0vTIz9-RG4N0kyz6MAjYp8FKCoF2i1AjsdJSC0NRRWwQ8sznvbc6zug0Hsi2HNyRntfp69whxHMcJyimcVHnFRF1nZdYgKNdOA3bypEaCGzyh0MrxvVzEcc183rtb8B1-Gi6JOR9UBApAwPMyC9g52DeHFOHiVgZCNTQjCoA9RicPSXnCJInWtb1pmFISugkv8hRMa7Xh1jyrASCxCkAFFLRgJ2QCMXEqNzFnTvCSw6DvBmoBVx1EbQTWPztA2o54f6owqWnwZ9k6nTgABxWivK3K7Pygm9Q++IZaKcHCTFIw2KKIpmU-PYEIDKbOkfPOuAAEAPogqAY96Mk5ze7m2YBvTHjwHe78ThwZNjZEskZKFB5dkGt4ETxt9tPF75H9H0YGBZkMv9iaDUXXzLJWYApJ6XxRptHVZJfOV43f99-F6Fxj-CoBBknUFHnvgfJqDcGx066bwatvewL9D6AJ-ifM+aUm7Wxvmje+fIACSqAoFv3Jh-ci38lzpG7kDGMMCgxTyaKsGedBNreljl-Uhs5WigAQMtPoq8AxoHSKgRgIBpiMw1KghqnAQzcN4fwx02xh4iJ4XwqAxsKGm3gDQ3W9C4ywOEhqQYqAojpEiNomulkUhpASFTAoTgMDuDhGg3EC0KhVBqDkZOkUh5lAnFIKAIYnauGgKYyMY9E4TycVOXGBjEQgFFAgUuPBcGfgMUPWQYwokeJiXaAxbBrjLRSTwfBjDmHLXkZQIAA):

```ts
type TCPDial = { signal?: AbortSignal }
export class TCPTransport implements Transport<TCPDial> {
//            ^^ Class 'TCPTransport' incorrectly implements interface 'Transport<TCPDial>'.
  constructor(upgrader: Upgrader, ...others: any) {
    return this
  }
  /**
   * Dial a given multiaddr.
   */
  dial(ma: Multiaddr, options?: TCPDial): Promise<Connection> {
    throw 1
  }
  /**
   * Create transport listeners.
   */
  createListener(options: any, handler: (connection:Connection) => void): Listener {
    throw 2
  }
  /**
   * Takes a list of `Multiaddr`s and returns only valid addresses for the transport
   */
  filter(multiaddrs: Multiaddr[]): Multiaddr[] {
    return multiaddrs
  }
}
```

P.S.: It complains about the prototype field but if you remove prototype from interface it will still complain about `new`.

src/transport/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Dial a given multiaddr.
+   */
+  dial(ma: Multiaddr, options?: DialOptions): Promise<Connection>;
+  /**
+   * Create transport listeners.
+   */
+  createListener(options: any, handler: (Connection) => void): Listener;
```
Gozala, line 23: looks like `options` here should be also made into generic rather than any.

src/transport/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  open: number;
+  upgraded?: number;
+  close?: number;
+}
+
+export type MultiaddrConnection = {
+  sink: Sink;
+  source: () => AsyncIterable<Uint8Array>;
+  close: (err?: Error) => Promise<void>;
+  conn: any;
```
Gozala, line 69: I really don't like any here, I think it's best to type it or remove it or at least type as `conn: unknown` and make the consumer of the `MultiaddrConnection` check what is it rather than treat it as anything.

src/transport/types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export type DialOptions = {
+  signal?: AbortSignal
+}
+
+/**
+ * A libp2p transport is understood as something that offers a dial and listen interface to establish connections.
+ */
+export interface Transport <DialOptions extends { signal?: AbortSignal }> {
+  new (upgrader: Upgrader, ...others: any): Transport<DialOptions>; // eslint-disable-line
```
Gozala, line 14: `...others: any` here also seems like bad idea, it should probably be either typed or made generic.

src/transport/types.ts
Gozala, line 23: Type check fails without this:

```suggestion
  createListener(options: any, handler: (connection: Connection) => void): Listener;
```

vasco-santos: good catch

vasco-santos: I can change it to unknown, but really not much more. 
Taking a look at https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/transport#create-a-listener there is no interface options for the listener. Each transport will have their own.

src/pubsub/index.js
vasco-santos: yeah, this seems a good call

src/pubsub/message/sign.js
```diff
@@ -31,24 +31,28 @@ async function signMessage (peerId, message) {
...
   // Get message sans the signature
   const baseMessage = { ...message }
   delete baseMessage.signature
   delete baseMessage.key
-  baseMessage.from = PeerId.createFromCID(baseMessage.from).toBytes()
+
   const bytes = uint8ArrayConcat([
     SignPrefix,
-    Message.encode(baseMessage)
+    Message.encode(Object.assign(baseMessage, {
```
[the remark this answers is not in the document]
vasco-santos: much cleaner, changing it.

src/pubsub/index.js
```diff
@@ -21,6 +21,18 @@ const {
...
+ * @typedef {import('../connection/connection')} Connection
+ * @typedef {import('./message').RPC} RPC
+ * @typedef {import('./message').SubOpts} RPCSubOpts
+ * @typedef {import('./message').Message} RPCMessage
+ * @typedef {import('./signature-policy').SignaturePolicyType} SignaturePolicyType
+ */
+
 /**
  * @typedef {Object} InMessage
  * @property {string} [from]
```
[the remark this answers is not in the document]
vasco-santos: It is really dependent on the signature policy. Have a look into the `validate` function. The same for other places like the `verifySignature` function, which `validate` only calls if it has a signature. I will add an error there to avoid potential future usages of the function though

src/pubsub/peer-streams.js
vasco-santos: I changed the if clause to include the `outBound` stream and avoid the `ts-ignore`.
I do not think we should change all the internal codebase to fulfil every single type scenario, otherwise we would just do typescript. For the user based APIs I completely agree that we should make this effort, but `peer-streams` is an internal class for pubsub routers

src/pubsub/peer-streams.js
vasco-santos: We need it for the check in the end of the function to only emit `stream:outbound` if the connection is new. We should only emit it when the outbound stream is ready, so this should be in the end, which means we need that auxiliary constant

src/transport/types.ts
hugomrdias: Yes, here we don't have any implementation just the interface, so if a method receives Transport reference and create an instance the type needs to be constructable

vasco-santos: FYI I am trying out @Gozala's suggestion with the factory below. It seems a better solution for this

hugomrdias:
Ts check is just commented out in the config because it's fail if a repo doesn't have the tsconfig (looking at you @Gozala this could be better 🙏🏼)
Regarding the transport issue I don't really understand the dial suggestion but seems that at least the prototype and new should go into a separate type *Factory or *Construtor something like that

vasco-santos:
So, I did a new iteration on this with the latest feedback from @Gozala 👍 :
- Record is now just an interface
- Transport and StreamMuxer now have a Factory interface for the constructor
- Pubsub minor fixes

This should now be ready to go.

.github/workflows/main.yml
```diff
@@ -0,0 +1,53 @@
...
+      - master
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: yarn
+    - run: yarn lint
+    # - uses: gozala/typescript-error-reporter-action@v1.0.4
```
hugomrdias, line 17:
```suggestion
    - uses: gozala/typescript-error-reporter-action@v1.0.4
```

hugomrdias (approved):

src/transport/types.ts
Gozala: But that also the whole point of generics to define a generic transport interface so that specific ones can specify concrete types that correspond to them just like it was in the case of dial:

```ts
interface Tansport<DialOptions extends { signal: AbortSignal }, ListenerOption> {
   dial(ma: Multiaddr, options?: DialOptions): Promise<Connection>;
   createListener(options: ListenerOptions, handler: (Connection) => void): Listener;
    filter(multiaddrs: Multiaddr[]): Multiaddr[];
}

class TCPTransport implements Transport<TCPDialOptions, TCPListenerOptions> {
   dial(ma: Multiaddr, options?: DialOptions): Promise<Connection> {
     // ...
   }
   createListener(options: TCPListenerOptions, handler: (connection: Connection) => void): Listener {
      // ...
   }
   // ...
}

type TCPDialOptions = {
  // ...
}

type TCPListenerOptions = //...
```

Gozala:
> Ts check is just commented out in the config because it's fail if a repo doesn't have the tsconfig (looking at you @Gozala this could be better 🙏🏼)

There is `tsconfig.json` so can't we enable it ? I'm sorry but I really don't want to deal with repos that do not have `tsconfig.json` because it needs to report same errors as running `tsc` locally. If we don't want to have a `tsconfig.json` in repo we could just run `aegir ts --preset config > tsconfig.json ` prior to running the checker.

Gozala:
> Regarding the transport issue I don't really understand the dial suggestion but seems that at least the prototype and new should go into a separate type *Factory or *Construtor something like that

Have you seen this comment https://github.com/libp2p/js-libp2p-interfaces/pull/74#issuecomment-737540458 it contains links to the TS playground illustrating issues. IMO whole point of the interface is to be able to tell in the class implementation that it **implementns** some interface. Current version is not implementable (or more like it's not doing what I think assumbtion is about it). There is also a second link that illustrates how it could be split into two interfaces such that class can say it implements Transport and the other `TrasportFactory` (or whatever you want to call it) that can be used in places that need a thing that instantiates a `Transport`.

P.S.: `prototype` field isn't required by either interface.

Gozala (approved):

hugomrdias (approved):

vasco-santos:
Per https://github.com/libp2p/js-libp2p/pull/802 feedback, I added two new minor commits to fix review issues

PR #75, opened 2020-12-02
Feat/add types 2

Gozala:
I end up fixing bunch of issues that were appearing in #74 when type-checking with new aegir.

### DIFF
--- a/src/connection/status.js
+++ b/src/connection/status.js
@@ -1,7 +1,12 @@
 'use strict'
 
-module.exports = {
+const STATUS = {
   OPEN: /** @type {'open'} */('open'),
   CLOSING: /** @type {'closing'} */('closing'),
   CLOSED: /** @type {'closed'} */('closed')
 }
+module.exports = STATUS
+
+/**
+ * @typedef {STATUS[keyof STATUS]} Status
+ */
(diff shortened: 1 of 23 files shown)


src/connection/status.js
```diff
@@ -1,7 +1,12 @@
 'use strict'
 
-module.exports = {
+const STATUS = {
```
vasco-santos, line 3: This seems to be causing issues integrating with libp2p PR

vasco-santos: "Property 'OPEN' does not exist on type 'typeof import("/Users/vsantos/work/pl/gh/libp2p/js-libp2p-interfaces/dist/src/connection/status")'"

Using: 
```js
const ConnectionStatus = require('libp2p-interfaces/src/connection/status')

ConnectionStatus.OPEN
```

src/pubsub/index.js
```diff
@@ -243,6 +243,7 @@ class PubsubBaseProtocol extends EventEmitter {
     try {
       const { stream, protocol } = await conn.newStream(this.multicodecs)
       const peer = this._addPeer(peerId, protocol)
+      // @ts-ignore MuxedStream is not DuplexIterableStream
```
vasco-santos, line 246: I will address this after merging. This should really be using the muxedStream

vasco-santos (approved):
Thanks, this looks good!
I am merging it into the other PR to test it against libp2p and fix the connection status + pubsub stream in there

PR #79, opened 2021-01-01
feat: add raw connection and transport tag to Connection interface

HexaField:
Adds the ability to optionally expose the raw connection of a transport as well as to provide a transport identifier tag in the Connection interface.

Changes:

`MultiaddrConnection` type now has optional `transportTag` parameter
`Connection` class now has optional `rawConn` and `transportTag` parameters

Solves in part https://github.com/libp2p/js-libp2p/issues/838

### DIFF
--- a/src/connection/README.md
+++ b/src/connection/README.md
@@ -48,6 +48,7 @@ A valid connection (one that follows this abstraction), must implement the follo
 - type: `Connection`
 ```js
 new Connection({
+  rawConn,
   localAddr,
   remoteAddr,
   localPeer,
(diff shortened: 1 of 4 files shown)


vasco-santos (changes requested):
@HexaField sorry for the late review. Let's get to add this feature now.

Per https://github.com/libp2p/js-libp2p/issues/838#issuecomment-766799676 we should expose the [multiaddrConn](https://github.com/libp2p/js-libp2p-interfaces/blob/e2419ea308b5db38966850ba6349602c93ce3b0e/src/transport/types.ts#L63) which will include the raw connection underneath.

jacobheun:
For the connection we should just expose the multiaddr connection instead of its underlying raw connection. I don't think we need the transportTag, we should be able to inspect this via the multiaddresses.

lidel:
Closing this PR as we'd like to just expose the underlying connection per https://github.com/libp2p/js-libp2p/issues/838#issuecomment-766799676

Issue #58, opened 2020-08-07
Split the tests or interfaces out?


achingbrain:
`ipfs-bitswap` has a runtime dependency on this module to pull in `MulticodecTopology` which means it ends up installing chai and sinon and whole bunch of other testing related modules.

This module depends on `libp2p-tcp` to run a test, and `libp2p-tcp` has a dev dependency on this module to pull in the tests.

This is a bit of a code smell, maybe either the tests or the interfaces should be in a separate module to break these weird dependency loops?

vasco-santos:
Splitting makes sense to me! I would go with a new repo/module for the tests: `js-libp2p-interfaces-compliance`

What is your opinion @jacobheun ?

jacobheun:
We should definitely avoid the test modules getting installed when base interfaces are being installed. I'd like to consider the usage more before just separating the two. Unless there is a strong need to get this fixed, let's revisit this in a couple months.

BigLep:
@vasco-santos : what are the next steps on this issue?  If you're paused working on it, do you want to return to weekly candidates and denote what's next please?

vasco-santos:
#97 has in the OP the remaining work. Basically handling the release of the multiple modules by updating the release scripts.

PR #81, opened 2021-01-21
chore: complete pubsub tests

vasco-santos:
Context: https://github.com/libp2p/js-libp2p/issues/857 let's remove peerDeps and get node16 in libp2p CI.

This basically [grabs](https://github.com/libp2p/js-libp2p/pull/866) the pubsub subsystem tests from `js-libp2p` and put them here

This PR restructures the pubsub tests to not use the routers as devDependencies in the libp2p repo, delegating the ownership of the integration testing to the modules. As a result, this PR moves the pubsub core tests from libp2p to the interface. This way, routers will run them as compliance tests with libp2p.

Discussion point/suggestion:
- Perhaps libp2p interface tests should start receiving an instance of libp2p configured with the module to test, instead of the module instance. This would allow us to create the test using libp2p API instead of the module, which for example to dial other peers mean accessing the libp2p instance within the module context.

Unblocks:

- [ ] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/148
- [ ] https://github.com/libp2p/js-libp2p-floodsub/pull/113
- [ ] https://github.com/libp2p/js-libp2p/pull/866


vasco-santos:
@achingbrain let me know your thoughts. Specially, if we should keep the same interface testing behaviour of receiving the libp2p module instance for testing, or if we should change all the interface tests to receive the libp2p instance. This can also be a follow up PR to not break the consistency with the other interfaces now

achingbrain:
> Perhaps libp2p interface tests should receive an instance of libp2p instead of the module instance

My gut feeling here is that this is not a good idea as it turns interface tests into integration tests which may hinder our ability to completely exercise the interface during the test, if some aspects of the interface require certain setup or other conditions.

achingbrain:
Unrelated but part of me thinks we should split the interface tests from the interfaces themselves, otherwise you end up with aegir being an undeclared dep of code that can be required when running in production.

vasco-santos:
> My gut feeling here is that this is not a good idea as it turns interface tests into integration tests which may hinder our ability to completely exercise the interface during the test, if some aspects of the interface require certain setup or other conditions.

Yeah, I agree let's wait for testground for more complex integration tests.

The PR as it is now, moves the pubsub subsystem tests from libp2p to the interface, which is how libp2p expects the routers to work and to have consistent behaviours. So, while they were technically used as integration tests in libp2p, they really are interface tests.

> Unrelated but part of me thinks we should split the interface tests from the interfaces themselves, otherwise you end up with aegir being an undeclared dep of code that can be required when running in production.

I have the same opinion. It is super weird to have the tests in the `src` folder but this is the result of past decisions. You already opened https://github.com/libp2p/js-libp2p-interfaces/issues/58 in the past. I can also promote it to the maintenance weekly candidates of libp2p.

achingbrain (approved):

Issue #85, opened 2021-03-11
findProviders should return PeerIDs


vasco-santos:
@Gozala commented on [Wed Mar 10 2021](https://github.com/libp2p/js-libp2p/issues/899)

findProviders returns `{ id: PeerId, multiaddrs: Multiaddr[] }`

https://github.com/libp2p/js-libp2p/blob/8e1fc78353cba194c86102dafcd68373b3f9a796/src/content-routing/index.js#L43-L52

While bitswap seems to assume it returns `PeerId`s instead. For more context please see

https://github.com/ipfs/js-ipfs-bitswap/pull/261/files#r587862380

---

@vasco-santos commented on [Thu Mar 11 2021](https://github.com/libp2p/js-libp2p/issues/899#issuecomment-796596628)

The link does not properly route me to the correct place I think.

This returns an async iterable if `{{ id: PeerId, multiaddrs: Multiaddr[] }`. In bitswap it seems to return `@returns {AsyncIterable<Provider>}` where Provider is:

```
 * @typedef {Object} Provider
 * @property {PeerId} id
 * @property {Multiaddr[]} multiaddrs
```

I did not understand the inconsistency here. Let me know what I am missing

---

@Gozala commented on [Thu Mar 11 2021](https://github.com/libp2p/js-libp2p/issues/899#issuecomment-796898855)

Here is the snapshot of the conversation in that pull 

![image](https://user-images.githubusercontent.com/21236/110823974-73155980-8247-11eb-83d3-e5b018fe0c51.png)

As per @achingbrain comment

> `libp2p.contentRouting.findProviders` should return only PeerIDs.
> 
> The idea is that libp2p manages the actual multiaddrs internally and we don't worry about them at this level.

Maybe it's all correct, because types seem to align now https://github.com/ipfs/js-ipfs-bitswap/pull/304

---

@vasco-santos commented on [Thu Mar 11 2021](https://github.com/libp2p/js-libp2p/issues/899#issuecomment-796906719)

I did not see that comment. So, you mean that `findProviders` should return `{AsyncIterable<{peerId: PeerId>}`

Ok, so @achingbrain is right in the sense that libp2p should handle internally the multiaddrs storage in the AddressBook, which is what we currently do.

[...]

PR #87, opened 2021-03-24
chore: update types

vasco-santos:
This PR is a follow up PR for types support. It includes:

- `aegir` update to latest, with new types problems fixed and `ts` files renamed to `d.ts`
- pubsub with options types available
- typedefs for remaining interfaces: discovery and routing


src/topology/index.js
```diff
@@ -62,6 +62,9 @@ class Topology {
     return Boolean(other && other[topologySymbol])
   }
 
+  /**
+   * @param {any} registrar
```
vasco-santos, line 66: We would need to create a circular dependency to have registrar types in the interface. We can change this interaction, but I feel we should not spend a lof of time reinventing this interaction at the moment.

src/peer-routing/types.d.ts
```diff
@@ -0,0 +1,8 @@
+import PeerId from 'peer-id'
+import Multiaddr from 'multiaddr'
+
+declare class PeerRouting {
```
hugomrdias, line 4: its not exported ?

src/peer-discovery/tests/index.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
hugomrdias, line 1: why the no check ?

hugomrdias (changes requested):

src/peer-routing/types.d.ts
vasco-santos: I was doing a named import on `libp2p`, but it makes sense to export this by default. I changed it here and in the other relevant types files

src/peer-discovery/tests/index.js
vasco-santos: I am excluding the interface tests for now. This is not a priority and we should probably tackle these together with the actual modules implementations to have type checks there

hugomrdias (approved):

package.json
```diff
@@ -47,39 +47,43 @@
   },
   "homepage": "https://github.com/libp2p/js-interfaces#readme",
   "dependencies": {
-    "@types/bl": "^2.1.0",
+    "@types/bl": "4.1.0",
```
rvagg, line 50: has this lost the `^` for a reason?

vasco-santos: good catch, mistyped! I will PR this shortly

src/transport/types.d.ts
```diff
@@ -62,7 +63,7 @@ export type MultiaddrConnectionTimeline = {
 
 export type MultiaddrConnection = {
   sink: Sink;
-  source: () => AsyncIterable<Uint8Array>;
+  source: AsyncIterable<Uint8Array | BufferList>;
```
rvagg, line 66: What are the use-cases for returning a plain `BufferList`, are we doing this as an optimisation to avoid buffer concatenation? Do our code paths actually take advantage of this or do we end up concatenating anyway?

src/transport/types.d.ts
```diff
@@ -1,3 +1,4 @@
+import BufferList from 'bl'
```
rvagg, line 1: So when you import `'bl'` but have `@types/bl` then it automagically pulls those definitions in rather than looking for the real `bl` (which isn't a dependency)?

package.json
```diff
@@ -47,39 +47,43 @@
...
     "protons": "^2.0.0",
-    "sinon": "^9.2.4",
+    "sinon": "^10.0.0",
     "streaming-iterables": "^5.0.4",
-    "uint8arrays": "^2.0.5"
+    "uint8arrays": "^2.1.3"
   },
   "devDependencies": {
-    "aegir": "^29.2.0",
+    "@types/debug": "^4.1.5",
```
rvagg, line 80: I'm assuming this is a devdep because it's an internal concern and we only want it to typecheck internal consistency and don't expose `debug` to users in any way form here. Is that correct?

src/content-routing/types.d.ts
```diff
@@ -0,0 +1,11 @@
+export = ContentRouting;
+
+import PeerId from 'peer-id'
+import Multiaddr from 'multiaddr'
+import CID from 'cids'
+
+declare class ContentRouting {
+  constructor (options: Object);
+  provide (cid: CID): Promise<void>;
+  findProviders (cid: CID, options: Object): AsyncIterable<{ id: PeerId, multiaddrs: Multiaddr[] }>;
```
rvagg, line 10: maybe better to define a `Provider` here rather than an anonymous type for `{id, multiaddrs}`?

src/peer-routing/types.d.ts
```diff
@@ -0,0 +1,10 @@
+export = PeerRouting;
+
+import PeerId from 'peer-id'
+import Multiaddr from 'multiaddr'
+
+declare class PeerRouting {
+  constructor (options?: Object);
+  findPeer (peerId: PeerId, options?: Object): Promise<{ id: PeerId, multiaddrs: Multiaddr[] }>;
```
rvagg, line 8: the `{id,multiaddrs}` appears twice here again, further suggests it should be it's own `Provider` type

src/pubsub/index.js
```diff
@@ -589,7 +595,7 @@ class PubsubBaseProtocol extends EventEmitter {
     for (const topic of message.topicIDs) {
       const validatorFn = this.topicValidators.get(topic)
       if (!validatorFn) {
-        continue
+        continue // eslint-disable-line
```
rvagg, line 598: you could just flip the logic here to `if (validatorFn) { await validatorFn(..) }` to avoid the disable

src/pubsub/message/index.js
```diff
@@ -1,17 +1,16 @@
...
-/**
- * @module pubsub/message/index
- */
-exports = module.exports
-exports.rpc = rpcProto
-exports.td = topicDescriptorProto
-exports.RPC = RPC
-exports.Message = RPC.Message
-exports.SubOpts = RPC.SubOpts
+module.exports = {
```
rvagg, line 10: out of interest: was this change forced by the type checking?

src/transport/types.d.ts
vasco-santos: yeah, if you are in typescript it will look into `@types` namespace. You can also do this in jsdoc for specifying typedefs with import:

```
@typedef {import('bl')} BufferList
```

package.json
vasco-santos: yes, that is correct!

src/transport/types.d.ts
vasco-santos: We use `BufferList` because some modules we use (example: https://github.com/bustle/streaming-iterables) also support them and it is valid from users stand point, when they create libp2p protocols. Anyway, we don't use this internally anyway for optimisation and we end up concatenating as you mention. 
This is something that we can look into

src/pubsub/message/index.js
vasco-santos: By the linter, but I think this can cause problems with inferred types and some bundlers. @hugomrdias has the bigger picture here on why we should avoid this

src/content-routing/types.d.ts
vasco-santos: This is more a representation of a peer rather then a Provider. As you pointed below, findPeer also has the same. We have been discussing about potentially change the API return type of this in https://github.com/libp2p/js-libp2p-interfaces/issues/85

We will probably look back into this

src/pubsub/index.js
vasco-santos: I will also update this in the follow up PR

PR #89, opened 2021-04-07
chore: update pubsub interface, multiaddr and remove protons

vasco-santos:
This PR adds some minor tweaks following up from feedback on #87 and updates multiaddr

Moreover, updating all the dependencies and removing protons in favour of protobufjs for smaller bundles and avoiding bundle problems with `web-encoding`

Most of the changes are auto generated files

Needs:

- [x] https://github.com/alanshaw/it-length-prefixed/pull/15
- [ ] https://github.com/alanshaw/it-length-prefixed/pull/16

Unblocks:

- [ ] https://github.com/libp2p/js-libp2p-floodsub/pull/114
- [ ] https://github.com/ChainSafe/js-libp2p-gossipsub/pull/153


package.json
```diff
@@ -47,7 +54,7 @@
   },
   "homepage": "https://github.com/libp2p/js-interfaces#readme",
   "dependencies": {
-    "@types/bl": "4.1.0",
+    "@types/bl": "^4.1.0",
```
rvagg, line 57: probably should be "^5.0.0" now, Alex updated it and it was released yesterday, bl@5 has been out for a week or so too thanks to Alex as well

src/pubsub/index.js
```diff
@@ -396,7 +396,7 @@ class PubsubBaseProtocol extends EventEmitter {
     if (msgs.length) {
       // @ts-ignore RPC message is modified
       msgs.forEach((message) => {
-        if (!(this.canRelayMessage || message.topicIDs.some((/** @type {string} */ topic) => this.subscriptions.has(topic)))) {
+        if (!(this.canRelayMessage || (message.topicIDs && message.topicIDs.some((topic) => this.subscriptions.has(topic))))) {
```
rvagg, line 399: oo, was this a bug that types picked up?

src/pubsub/message/rpc.d.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        /**
+         * Converts this SubOpts to JSON.
+         * @returns JSON object
+         */
+        public toJSON(): { [k: string]: any };
+    }
+
+    /** Properties of a Message. */
+    interface IMessage {
```
rvagg, line 135: why are the properties here all both optional and allow `null`?

rvagg (approved):
sgtm (didn't review the two generated files, I'm assuming there's not much to say there)

The only major thing that stands out is the liberal use of `?` combined with `|null` in the types. I suppose if this is just an initial go at external facing interfaces then that might an ok start, but it seems to suggest some looseness that maybe could be tightened up later?

src/pubsub/index.js
vasco-santos: This is because the protobuf definition does not state topicIDs as [required](https://github.com/libp2p/js-libp2p-interfaces/blob/v0.9.0/src/pubsub/message/rpc.proto.js#L16). As far as I know, all implementations send an empty arrau and not undefined and this has not been a problem. But could be as it is not spec'ed as required

src/pubsub/message/rpc.d.ts
vasco-santos: This is generated from the protobuf [spec](https://github.com/libp2p/specs/tree/master/pubsub#the-message) for pubsub where no part of the message is required. But basically, "Depending on the signature policy and message ID function."

vasco-santos:
> The only major thing that stands out is the liberal use of ? combined with |null in the types. I suppose if this is just an initial go at external facing interfaces then that might an ok start, but it seems to suggest some looseness that maybe could be tightened up later?

As we reuse the same definition for several types of pubsub messages supported, the protobuf states them as optional, which makes them possibly null. More details in review comments

package.json
achingbrain: bl seems like a weird thing to use at the interface level, perhaps it's something we can pull out in future?

It's only used [in one place](https://github.com/libp2p/js-libp2p-interfaces/blob/master/src/pubsub/index.js#L348) and from the type def it looks like it's only necessary because the exported BufferList can't be treated as a Uint8Array based on the type heirachy.

src/pubsub/message/topic-descriptor.proto
```diff
@@ -1,10 +1,10 @@
-'use strict'
-module.exports = `
+syntax = "proto3";
+
 // topicCID = cid(merkledag_protobuf(topicDescriptor)); (not the topic.name)
 message TopicDescriptor {
   optional string name = 1;
   optional AuthOpts auth = 2;
-  optional EncOpts enc = 2;
+  optional EncOpts enc = 3;
```
achingbrain, line 7: Oops.

src/pubsub/peer-streams.js
```diff
@@ -9,6 +9,7 @@ const log = Object.assign(debug('libp2p-pubsub:peer-streams'), {
 /** @type Events */
 const EventEmitter = require('events')
 
+// @ts-ignore TODO: https://github.com/alanshaw/it-length-prefixed/pull/15
```
achingbrain, line 12: This has been merged, can be removed?

vasco-santos: I am waiting on releasing this: https://github.com/alanshaw/it-length-prefixed/pull/16
Otherwise tests fail

src/pubsub/message/topic-descriptor.proto
vasco-santos: yes, this was awkward! Go and the spec were right, but I guess there is no one using topic descriptors in JS so far
Spec: https://github.com/libp2p/specs/tree/master/pubsub#the-topic-descriptor
Go: https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L61

package.json
```diff
@@ -57,28 +64,28 @@
...
     "streaming-iterables": "^5.0.4",
     "uint8arrays": "^2.1.3"
   },
   "devDependencies": {
     "@types/debug": "^4.1.5",
-    "aegir": "^32.1.0",
+    "aegir": "^33.0.0",
     "cids": "^1.1.6",
     "events": "^3.3.0",
     "it-handshake": "^1.0.2",
```
achingbrain, line 91:
```suggestion
    "it-handshake": "^2.0.0",
```

src/pubsub/message/topic-descriptor.proto
vasco-santos: Types catching all the bugs 🎉

PR #90, opened 2021-04-19
feat: muxed stream close read and write

vasco-santos:
__ Created new PR with changes from #67 on top of current master branch __

This adds tests for closeRead and closeWrite on the muxer streams.

Also:

- Removed libp2p-tcp in favor of it-pair to avoid circular deps
- Connection.close() will now close its internal streams to avoid lingering streams.

BREAKING CHANGE: This adds closeWrite and closeRead checks in the tests, which will cause test failures for muxers that dont implement those.

Closes #67

### DIFF
--- a/src/connection/connection.js
+++ b/src/connection/connection.js
@@ -220,6 +220,8 @@ class Connection {
 
     this.stat.status = CLOSING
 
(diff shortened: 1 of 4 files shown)


Gozala (approved):

achingbrain:
Superseded by #214

PR #94, opened 2021-04-21
chore: add dht selectors and validators

vasco-santos:
Adds dht selectors and validators here to be used across the board easily. 

Also removes unecessary EventEmitter types, as this was already fixed

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -1,2 +1,5 @@
-export type DhtSelectors = { [key: string]: function (Uint8Array, Uint8Array[]): number }
-export type DhtValidators = { [key: string]: { func: (key: Uint8Array, value: Uint8Array) => Promise<void> } }
+export type SelectFn = function (Uint8Array, Uint8Array[]): number
+export type ValidateFn = function (Uint8Array, Uint8Array): Promise<void>
+
+export type DhtSelectors = { [key: string]: SelectFn }
+export type DhtValidators = { [key: string]: { func: ValidateFn } }
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -1,18 +1,2 @@
-export interface EventEmitterFactory {
-  new(): EventEmitter;
-}
-
-export interface EventEmitter {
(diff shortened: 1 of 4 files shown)


src/types.d.ts
```diff
@@ -1,18 +1,5 @@
-export interface EventEmitterFactory {
-  new(): EventEmitter;
-}
+export type SelectFn = function (Uint8Array, Uint8Array[]): number
+export type ValidateFn = function (Uint8Array, Uint8Array): Promise<void>
```
achingbrain, line 2: If you add a typedef to `/src/index.js` like:

```js
/**
 * @typedef {import('./types').SelectFn} SelectFn
 * @typedef {import('./types').ValidateFn} ValidateFn
 * @typedef {import('./types').DhtSelectors} DhtSelectors
 * @typedef {import('./types').DhtValidators} DhtValidators
 */
```

then you can use them from other modules like:

```js
/**
 * @typedef {import('libp2p-interfaces').SelectFn} SelectFn
 */
```

otherwise you have to do:

```js
/**
 * @typedef {import('libp2p-interfaces/src/types').SelectFn} SelectFn
 */
```

Which is ok, but needs the `typesVersions` hack and prevents you from moving that file in future.

achingbrain (approved):
Small suggestion inline, not a blocker

src/types.d.ts
vasco-santos: This was causing:

```
src/validator.js:15:19 - error TS2306: File '/Users/vsantos/work/pl/gh/libp2p/js-libp2p-record/node_modules/libp2p-interfaces/dist/src/index.d.ts' is not a module.
15  * @param {import('libp2p-interfaces').DhtValidators} validators
```

After generated

```
type SelectFn = import('./types').SelectFn;
type ValidateFn = import('./types').ValidateFn;
type DhtSelectors = import('./types').DhtSelectors;
type DhtValidators = import('./types').DhtValidators;
//# sourceMappingURL=index.d.ts.map
```

I could not easily work around this issue, so I reverted to the original solution for now

PR #97, opened 2021-05-28
chore: monorepo separating interfaces and compliance tests

### LINKED ISSUE #58: Split the tests or interfaces out?

vasco-santos:
This PR makes libp2p-interfaces a monorepo with its interfaces and compliance tests in two different modules.

Closes #58

BREAKING CHANGE: the tests now live in the `libp2p-interfaces-compliance-tests` module


packages/compliance-tests/package.json
```diff
@@ -0,0 +1,86 @@
...
+    "abort-controller": "^3.0.0",
+    "abortable-iterator": "^3.0.0",
+    "aegir": "^33.2.2",
+    "chai": "^4.3.4",
+    "chai-checkmark": "^1.0.1",
+    "delay": "^5.0.0",
+    "it-goodbye": "^3.0.0",
+    "it-pair": "^1.0.0",
+    "it-pipe": "^1.1.0",
+    "libp2p-interfaces": "^0.11.0",
```
vasco-santos, line 46: @achingbrain how do you handle dependency updates of adjacent modules?

achingbrain: Do you mean when sibling versions change within a monorepo?  Lerna takes care of all that.

achingbrain (approved):

PR #98, opened 2021-06-02
chore: update to new multiformats

achingbrain:
Replaces the `cids`, `multihashes` and `multibase` modules with the
new multiformats module.

BREAKING CHANGE: uses the CID class from the new multiformats module and pubsub.getMsgId became async


src/pubsub/message/sign.js
```diff
@@ -44,13 +44,16 @@ async function verifySignature (message) {
     throw new Error('Message must contain a signature to be verified')
   }
 
+  if (!message.from) {
+    throw new Error('Message must contain a from to be verified')
```
vasco-santos, line 48:
```suggestion
    throw new Error('Message must contain a from property to be verified')
```

vasco-santos (commented):
Looks good, pending the final releases

vasco-santos:
Pointing out that there is a browser issue in CI

vasco-santos (approved):
LGTM

PR #100, opened 2021-06-28
draft: use interfaces instead of building blocks

Gozala:


lidel:
@vasco-santos thoughts on this? should conflicts be resolved, or need a new one?

vasco-santos:
I think the next step here is for me (or anyone else) look into the proposal and provide feedback. Then, we iterate and the conflicts are fixed

wemeetagain:
Closing since this repo is now solely interfaces and compliance tests

PR #103, opened 2021-07-07
fix: make tests more reliable

achingbrain:
await finishing processing rpc messages instead of using setTimeout.

Also return promises and await during processing message stream so we don't get unhandled promise rejections (potentially).


vasco-santos (approved):
LGTM

PR #104, opened 2021-07-08
fix: remove timeouts

achingbrain:
Control them with global --timeout settings instead as some CI systems have worse performance than others.


packages/compliance-tests/src/pubsub/connection-handlers.js
```diff
@@ -266,10 +263,8 @@ module.exports = (common) => {
 
         psA.on(topic, handlerSpy)
         psB.on(topic, handlerSpy)
-        await Promise.all([
-          psA.subscribe(topic),
-          psB.subscribe(topic)
-        ])
+        psA.subscribe(topic),
+        psB.subscribe(topic)
```
achingbrain, line 267: This is a synchronous operation so no need to wrap it in a promise.all

PR #118, opened 2021-12-15
Added events when peers connect or disconnect

jonaswre:
I think it would be usefull to emit peer connect and peer disconnect events. 

Currently libs  like pfs-pubsub-room do polling to generate these events. 
https://github.com/ipfs-shipyard/ipfs-pubsub-room/blob/8c00b4fbfa7cbd47889fe035dfa9ac23638d9b58/src/index.js#L34

But its far more efficient to just emit the events where they occur rather then polling them few steps up the stack.

I know there is already a event for when the subscription changes. But from what i've gathered that event is only emitted when the peer changes the subscription but not if a connect/disconnect happen. Connection is terminated without propper tear down.

### DIFF
--- a/packages/libp2p-interfaces/src/pubsub/index.ts
+++ b/packages/libp2p-interfaces/src/pubsub/index.ts
@@ -77,9 +77,18 @@ interface SubscriptionChangeEvent {
   peerId: PeerId
   subscriptions: Subscription[]
 }
+interface PeerConnectEvent {
+  peerId: PeerId
+}
+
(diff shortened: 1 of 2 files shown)


jonaswre:
@achingbrain is that something we could add? Its just a very tiny change but It would make client implementations bit easier.

jonaswre:
Anybody cares to comment on my change?

jonaswre:
Thanks for getting back to me. Its a big api. Sorry for missing that

PR #120, opened 2021-12-21
fix: make connection upgrade and encryption abortable

achingbrain:
Allow passing AbortSignals in to connection upgrading and encryption
so the operations can be aborted if a timeout is reached.


vasco-santos (approved):
LGTM

PR #107, opened 2021-09-24
feat: add PeerId interface and compliance tests

nazarhussain:
This PR will create `PeerId` interface and its compliance tests. The discussion was initiated from https://github.com/libp2p/js-libp2p/issues/955 and https://github.com/libp2p/js-peer-id/pull/150#pullrequestreview-692633354

PR #126, opened 2021-12-25
fix: peer-id privKey and pubKey can be undefined

nazarhussain:
In the PR https://github.com/libp2p/js-libp2p-interfaces/pull/107  the `PeerId` interface was introduced. Later in the PR https://github.com/libp2p/js-libp2p-interfaces/pull/111/ that was refactored and some intentional changes introduced. 

The `privKey` and `pubKey` both can be undefined depending on the data, more details can be found in the PR https://github.com/libp2p/js-peer-id/pull/150.

The `PeerId` object can be created with the following scenarios: 

1. `new` in that case both keys are optional 
2. `PeerId.create` in that case if provided input is valid than both keys exists 
3. `PeerId.createFromBytes`, `PeerId.createFromB58String`, `PeerId.createFromHexString`, `PeerId.createFromCID`  in these cases public key _can_ be decoded but private key don't exists 
4. `PeerId.createFromPubKey` in that case public key exists but not private 
5. `PeerId.createFromPrivKey` in that case both keys exist
6. `PeerId.createFromJSON` in that case both keys exist 
7. `PeerId.createFromProtobuf` in that case both keys exist

Considering all the above cases, both `privKey` and `pubKey` were initially set was set to have `undefined`  values. Seems the change in refactoring PR #111 was made unintended.

### DIFF
--- a/packages/compliance-tests/src/peer-id/index.js
+++ b/packages/compliance-tests/src/peer-id/index.js
@@ -378,7 +378,7 @@ module.exports = (common) => {
 
       it('missmatch private - public key', async () => {
         const digest = await k1.public.hash()
-        expect(factory.createFromJSON({
+        await expect(factory.createFromJSON({
           id: digest,
           pubKey: k1,
           privKey: k2.public
@@ -387,7 +387,7 @@ module.exports = (common) => {
 
       it('missmatch id - private - public key', async () => {
         const digest = await k1.public.hash()
-        expect(factory.createFromJSON({
+        await expect(factory.createFromJSON({
           id: digest,
           pubKey: k1,
           privKey: k3.public
@@ -395,7 +395,7 @@ module.exports = (common) => {
       })
 
       it('invalid id', () => {
-        expect(factory.createFromJSON('hello world')).eventually.be.rejectedWith(/invalid id/)
+        await expect(factory.createFromJSON('hello world')).eventually.be.rejectedWith(/invalid id/)
       })
     })
   })
--- a/packages/compliance-tests/src/peer-id/index.js
+++ b/packages/compliance-tests/src/peer-id/index.js
@@ -357,10 +357,6 @@ module.exports = (common) => {
       const peerId2 = factory.createFromB58String(peerId.toB58String())
 
       expect(peerId1).to.deep.equal(peerId2)
-
-      peerId1.toString()
-
-      expect(peerId1).to.deep.equal(peerId2)
     })
 
     describe('throws on inconsistent data', () => {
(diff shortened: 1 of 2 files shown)


achingbrain:
TLDR: please use the v2.x.x release for the time being, tagged `latest` on npm - PRs can target the `release/v2.x.x` branch.

----

Longer version:

The typescript version of libp2p-interfaces is trying to take the interface definitions in a new direction.  For peer id, please see the changes in https://github.com/libp2p/js-libp2p-interfaces/pull/117 - it makes peer ids essentially a place to store some `Uint8Array`s which you can then use to do encryption rather than pulling in `libp2p-crypto` by default.  This is to allow using peer id datatypes in lightweight clients like `ipfs-http-client` instead of having everything as opaque strings.

It's all a bit in flux and will be until https://github.com/libp2p/js-libp2p/issues/1021 lands.

nazarhussain:
@achingbrain  Please review.

packages/compliance-tests/src/peer-id/index.js
```diff
@@ -382,29 +378,24 @@ module.exports = (common) => {
...
       it('missmatch private - public key', async () => {
         const digest = await k1.public.hash()
-        expect(() => {
-          factory.createFromJSON({
-            id: digest,
-            pubKey: k1,
-            privKey: k2.public
-          }) // eslint-disable-line no-new
-        }).to.throw(/inconsistent arguments/)
+        expect(factory.createFromJSON({
```
achingbrain, line 381: You need to await on or return the promise returned when using `eventually`:

```suggestion
        await expect(factory.createFromJSON({
```

packages/compliance-tests/src/peer-id/index.js
```diff
@@ -382,29 +378,24 @@ module.exports = (common) => {
...
       it('missmatch id - private - public key', async () => {
         const digest = await k1.public.hash()
-        expect(() => {
-          factory.createFromJSON({
-            id: digest,
-            pubKey: k1,
-            privKey: k3.public
-          }) // eslint-disable-line no-new
-        }).to.throw(/inconsistent arguments/)
+        expect(factory.createFromJSON({
```
achingbrain, line 390: You need to await on or return the promise returned when using `eventually`:

```suggestion
        await expect(factory.createFromJSON({
```

packages/compliance-tests/src/peer-id/index.js
```diff
@@ -382,29 +378,24 @@ module.exports = (common) => {
...
+          id: digest,
+          pubKey: k1,
+          privKey: k3.public
+        })).eventually.be.rejectedWith(/inconsistent arguments/)
       })
 
       it('invalid id', () => {
-        // @ts-expect-error incorrect constructor arg type
-        expect(() => factory.createFromJSON('hello world')).to.throw(/invalid id/)
+        expect(factory.createFromJSON('hello world')).eventually.be.rejectedWith(/invalid id/)
```
achingbrain, line 398: You need to await on or return the promise returned when using `eventually`:

```suggestion
        await expect(factory.createFromJSON('hello world')).eventually.be.rejectedWith(/invalid id/)
```

achingbrain (changes requested):
Good catch - thanks! Just a few changes needed in order for the tests to not cause unhanded promise rejections if they fail.

Issue #114, opened 2021-11-30
Pubsub: Could not find a declaration file for module 'libp2p-pubsub'


wemeetagain:
Trying to update js-libp2p-gossipsub (tracking issue https://github.com/ChainSafe/js-libp2p-gossipsub/issues/171)

After installing `libp2p-pubsub` v0.7.0 and `libp2p-interfaces` v2.0.0
and adding this line:
```ts
import { PubsubBaseProtocol } from 'libp2p-pubsub'
```

I get the following error in my IDE:
```
Could not find a declaration file for module 'libp2p-pubsub'. '/home/cayman/Code/js-libp2p-gossipsub/node_modules/libp2p-pubsub/dist/src/index.js' implicitly has an 'any' type.
```

Also tried starting a new project and installing just those two packages, same error.

achingbrain:
Could you try again please?  There was a bit of churn and I think you got caught up in it.

`libp2p-interfaces@2.x.x` was the [TypeScript refactor](https://github.com/libp2p/js-libp2p/issues/1021) but it became necessary to release that as `v3` since `libp2p-crypto` removed support for node 14 and below.

`libp2p-interfaces@1.x.x` then became `libp2p-interfaces@2.x.x` for the same reason.

`libp2p-pubsub` is not ready for use yet, it needs more of the ecosystem porting to TS - please stick with the PubSub base class defined in `libp2p-interfaces@2.x.x` for now.

Sorry for the disruption.

lidel:
Closing as @wemeetagain say its no longer a problem :^)

PR #127, opened 2021-12-28
fix: pubsub typescript type export

wemeetagain:
Resolve https://github.com/libp2p/js-libp2p-interfaces/issues/114

~I'm not entirely sure what about the `typesVersions` field breaks the types, but... removing it resolves the problem.
We shouldn't need the `typesVersions` field anyways unless we have different types for different versions of typescript.~

Added an additional field to `typeVersions`. Based on how it's done in [`uint8arrays`](https://github.com/achingbrain/uint8arrays/blob/master/package.json#L43)

This probably should be applied to all of the packages, but I tested this locally against gossipsub.
LMK if you'd like me to apply this consistently.

### DIFF
--- a/packages/libp2p-pubsub/package.json
+++ b/packages/libp2p-pubsub/package.json
@@ -11,6 +11,7 @@
   "typesVersions": {
     "*": {
       "*": [
+        "dist/src",
         "dist/src/*",
         "dist/src/*/index"
       ]


achingbrain:
We need the `typesVersions` field to do deep requires, hence the ci failure:

```
Error: libp2p-interfaces-compliance-tests: src/pubsub/messages.ts(5,24): error TS2307: Cannot find module 'libp2p-pubsub/utils' or its corresponding type declarations.
Error: libp2p-interfaces-compliance-tests: src/pubsub/messages.ts(6,29): error TS2307: Cannot find module 'libp2p-pubsub/peer-streams' or its corresponding type declarations.
```

Chances are the types are breaking because `typesVersions` is missing a wildcard - if you build with [--traceResolution](https://www.typescriptlang.org/tsconfig/traceResolution.html) you should be able to see where tsc is trying to load the `.d.ts` from and why it's not doing what we think it should be doing.

wemeetagain:
Thanks for the `--traceResolution` advice, very helpful for debugging.

@achingbrain ready for another review

PR #128, opened 2021-12-28
feat: async peer store

achingbrain:
Refactors interfaces and classes used by `libp2p-interfaces` to use the async peer store from https://github.com/libp2p/js-libp2p/pull/1058

BREAKING CHANGE: peerstore methods are now all async

### DIFF
--- a/packages/compliance-tests/src/topology/multicodec-topology.js
+++ b/packages/compliance-tests/src/topology/multicodec-topology.js
@@ -5,9 +5,8 @@
 
 const { expect } = require('aegir/utils/chai')
 const sinon = require('sinon')
-
+const delay = require('delay')
 const PeerId = require('peer-id')
-
 const peers = require('../utils/peers')
 
 module.exports = (test) => {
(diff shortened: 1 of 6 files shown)


packages/compliance-tests/src/topology/multicodec-topology.js
```diff
@@ -58,6 +57,9 @@ module.exports = (test) => {
         protocols: Array.from(topology.multicodecs)
       })
 
+      // 'change:protocols' event handler is async
+      await delay(10)
```
vasco-santos, line 61: why not listen for the event?

vasco-santos (commented):
is looking good 🎉 
just left a nit on the tests

packages/compliance-tests/src/topology/multicodec-topology.js
achingbrain: The hander that invokes `topology._updatePeers` is async, so we can't just listen for the event as we're interested in the side-effects of the handler being invoked.

From a practical perspective the mock peer store used in the tests doesn't really do any async work, it just defers execution using the `async` keyword so `delay` here puts the rest of the test into the microtask queue - the handler is ahead of it in the same queue so this is enough for the test to pass.

Longer term we shouldn't test for side-effects like this as it's really fragile and makes refactoring hard.

PR #138, opened 2022-01-06
fix: partial revert of 131 for v2.x.x

achingbrain:
A PeerId's privKey and/or pubKey can be undefined but we've got types
that say they cannot.

These fields are accessed when a node is interacting with it's own
PeerId so these fields are set, but really they can be undefined.

Marking these fields as undefined was a breaking change so this PR
partially reverts #131 - the proper fix is in the 4.x.x release line
which should be used in preference to 2.x.x.


vasco-santos (approved):
thanks

PR #153, opened 2022-01-28
fix: remove extra fields

achingbrain:
Trying to cut the number of fields objects have down - some of these should be obvious from the context - our own peer id, etc.


wemeetagain (approved):
LGTM

PR #221, opened 2022-05-24
fix: dead badge

zeim839:
Fixes a typo that caused the [peer-discovery](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/libp2p-interfaces/src/peer-discovery) badge not to appear in Github previews. 

Please note that other readmes in the libp2p-interfaces folder point to their old, deprecated and read-only repos (example: [peer-routing](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/libp2p-interfaces/src/peer-routing)), whereas this one was set to point to the local images folder. Perhaps something should be done for consistency?


zeim839:
Badges now point to their local image folder. No sense in having one if you're not gonna use it!

wemeetagain (approved):
Thanks for the contribution @zeim839

Issue #226, opened 2022-06-09
Split non-interface packages out from this module


achingbrain:
Now that we've shipped the ts re-write, having so many non-interface modules in this monorepo is of dwindling use since we don't need to make big coordinated changes so much now.

All of our interfaces are in the one `libp2p-interfaces` module which is basically depended on by everything in the libp2p ecosystem, which means if we release a breaking change to one interface (say, the Registrar), we have to go and upgrade loads of unrelated repos (granted dependabot and CI releases make this less of a pain).

It would be better to move the non interface related modules out of this repo, and split the `libp2p-interfaces` module up into multiple modules (one per dir under [src](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/libp2p-interfaces/src) for example), and combine the each set of interfaces with it's test suite.  Then we can release breaking changes to an interface without having to upgrade everything, everywhere.

PR #228, opened 2022-06-10
chore: break modules apart

achingbrain:
Splits each interface area into it's own module so we can release breaking changes to (for example) the pubsub interface without also having to upgrade single transport module to the new `@libp2p/interfaces` major.

Moves non-interface/compliance tests packages into their own repos:

* https://github.com/libp2p/js-libp2p-components
* https://github.com/libp2p/js-libp2p-logger
* https://github.com/libp2p/js-libp2p-multistream-select
* https://github.com/libp2p/js-libp2p-peer-id
* https://github.com/libp2p/js-libp2p-peer-record
* https://github.com/libp2p/js-libp2p-peer-store
* https://github.com/libp2p/js-libp2p-pubsub
* https://github.com/libp2p/js-libp2p-toplogy
* https://github.com/libp2p/js-libp2p-tracked-map

Fixes: #226

BREAKING CHANGE: most modules have been split out of the `@libp2p/interfaces` and `@libp2p/interface-compliance-tests` packages


wemeetagain (approved):
Can you link to the new repos here?

achingbrain:
I've updated the comment

achingbrain:
I have deleted the npm token that will publish this change so the build should fail.  Will manually revert and merge #229 / #230 though likely tomorrow morning now.

wemeetagain:
sorry about that :(

achingbrain:
Not your fault.  #230 needs a small fix to ipfs that's in now before it can go out, but there was a delay because a) we actually go way over the new mplex per-connection stream limit during the tests and need the per-protocol limits to alleviate that and b) someone pushed a new github workflow PR that automerged to every ipfs repo and has eaten all the GH action runners for the org all day.

Perfect storm really 😆 😢 😭

PR #245, opened 2022-06-16
feat!: store stream data on the stream, track the stream direction

achingbrain:
* Adds a `.stat` property to `Stream`, similar to the one on `Connection`
* Moves the stream timeline to `.stat`, same as on `Connection`
* Stores stream metadata on the stream, instead of in a `.registry` on the `Connection` - it's not used by libp2p and storing streams/metadata separately is a memory leak waiting to happen
* Stores protocol information on the stream, instead of in `.registry` on the `Connection`
* Stores stream direction on the stream, instead of not at all
* Runs compliance tests on the mock connection class
* Protocol handlers do not receive the negotiated protocol name as part of the stream data, it's set on the stream now
* `newStream` on `Connection` returns a `Stream`
* `Connection` stats have a `ConnectionTimeline` and `Stream` stats have a `StreamTimeline` as the properties are different

BREAKING CHANGE: The Connection and Stream APIs have been updated

### DIFF
--- a/packages/interface-connection-compliance-tests/src/index.ts
+++ b/packages/interface-connection-compliance-tests/src/index.ts
@@ -114,7 +114,7 @@ export default (test: TestSetup<Connection>) => {
         expect(stream).to.have.nested.property('stat.protocol', protocol)
 
         // Close stream
-        await stream.close()
+        stream.close()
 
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
really big fan of this change :heart:

PR #253, opened 2022-06-20
feat: add direction to StreamMuxerInit

### LINKED ISSUE #250: Add direction to StreamMuxerInit
Yamux requires knowing ahead of time whether it is an inbound (server) or outbound (client) channel.
It uses this to select stream ids when it creates new streams.
inbound creates streams with even stream ids, outbound creates streams with odd stream ids.
If two yamux channels configured to both inbound or both outbound are connected, they will fatally close the channel at the first new incoming stream.

### LINKED ISSUE #252: Add direction type and/or enum
The direction type `'inbound' | 'outbound'` is repeated in many places.
It would be nice to have a `Direction` or `ConnectionDirection` type or enum to avoid this.

wemeetagain:
- add `Direction` type alias in interface-connection
- add `direction?: Direction` to `StreamMuxerInit`

Resolves #252 
Resolves #250

### DIFF
--- a/packages/interface-stream-muxer/src/index.ts
+++ b/packages/interface-stream-muxer/src/index.ts
@@ -1,5 +1,5 @@
 import type { Duplex } from 'it-stream-types'
-import type { Stream } from '@libp2p/interface-connection'
+import type { Direction, Stream } from '@libp2p/interface-connection'
 import type { AbortOptions } from '@libp2p/interfaces'
 
 export interface StreamMuxerFactory {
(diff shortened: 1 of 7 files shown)


packages/interface-stream-muxer/src/index.ts
```diff
@@ -24,4 +24,5 @@ export interface StreamMuxer extends Duplex<Uint8Array> {
 export interface StreamMuxerInit extends AbortOptions {
   onIncomingStream?: (stream: Stream) => void
   onStreamEnd?: (stream: Stream) => void
+  direction: Direction
```
achingbrain, line 27: Can this be optional?  Otherwise it's a breaking change if it's required.

achingbrain (changes requested):

PR #254, opened 2022-06-20
feat!: add stream muxer close

### LINKED ISSUE #249: Add muxer.close() to StreamMuxer
Yamux defines 'ping' and 'go away' messages that imply the ability to close the channel at the muxer level.
It would be nice to add a `close(): void` method on `StreamMuxer` so that libp2p can close the muxer when it should.

Yamux 'ping' allows for a 'keepalive' feature that pings on an interval. (Pings are used to determine RTT and used to in calculating whether to increase the received data window when promptly consuming data)
In order to properly clean up the long running keepalive task, `muxer.close()` must be called.
With `close` on `StreamMuxer`, we can enable keepalive by default.

Calling `muxer.close()` also lets the yamux send the 'go away' message, which signals to the other party that the channel is closed and will no longer be used.

wemeetagain:
BREAKING CHANGE: StreamMuxer now has a `close` method

Resolves libp2p/js-libp2p-interfaces#249

### DIFF
--- a/packages/interface-mocks/src/muxer.ts
+++ b/packages/interface-mocks/src/muxer.ts
@@ -266,6 +266,7 @@ class MockMuxer implements StreamMuxer {
   public name: string
   public protocol: string = '/mock-muxer/1.0.0'
 
+  private readonly closeController: AbortController
   private readonly registryInitiatorStreams: Map<string, MuxedStream>
   private readonly registryRecipientStreams: Map<string, MuxedStream>
   private readonly options: StreamMuxerInit
@@ -279,17 +280,11 @@ class MockMuxer implements StreamMuxer {
     this.registryRecipientStreams = new Map()
     this.log('create muxer')
     this.options = init ?? { direction: 'inbound' }
+    this.closeController = new AbortController()
     // receives data from the muxer at the other end of the stream
     this.source = this.input = pushable({
       onEnd: (err) => {
-        this.log('closing muxed streams')
-        for (const stream of this.streams) {
-          if (err == null) {
-            stream.close()
-          } else {
-            stream.abort(err)
-          }
-        }
+        this.close(err)
       }
     })
 
@@ -303,7 +298,7 @@ class MockMuxer implements StreamMuxer {
   async sink (source: Source<Uint8Array>) {
     try {
       await pipe(
-        source,
+        abortableSource(source, this.closeController.signal),
         (source) => map(source, buf => uint8ArrayToString(buf)),
         ndjson.parse,
         async (source) => {
@@ -366,6 +361,9 @@ class MockMuxer implements StreamMuxer {
   }
 
   newStream (name?: string) {
+    if (this.closeController.signal.aborted) {
+      throw new Error('Muxer already closed')
+    }
     this.log('newStream %s', name)
     const storedStream = this.createStream(name, 'initiator')
     this.registryInitiatorStreams.set(storedStream.stream.id, storedStream)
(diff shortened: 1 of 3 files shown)


packages/interface-stream-muxer-compliance-tests/src/close-test.ts
```diff
@@ -67,6 +67,48 @@ export default (common: TestSetup<StreamMuxerFactory>) => {
       expect(dialer.streams).to.have.lengthOf(0)
     })
 
+    it('calling close closes streams', async () => {
```
achingbrain, line 70: Needs a test where an error is passed to `.close` to verify the correct behaviour.

packages/interface-stream-muxer-compliance-tests/src/close-test.ts
```diff
@@ -67,6 +67,48 @@ export default (common: TestSetup<StreamMuxerFactory>) => {
...
+      let openedStreams = 0
+      const expectedStreams = 5
```
achingbrain, line 72: Also, if a muxer is closed, calling `newStream` should probably throw?  That would need a test too.

packages/interface-mocks/src/muxer.ts
```diff
@@ -399,6 +394,19 @@ class MockMuxer implements StreamMuxer {
...
+    if (this.closeController.signal.aborted) return
+    this.log('closing muxed streams')
+
+    if (err == null) {
+      this.streams.forEach(s => s.close())
+    } else {
+      this.streams.forEach(s => s.abort(err))
+    }
+    this.closeController.abort()
+    this.input.end(err)
```
achingbrain, line 408: Calling `this.input.end(err)` may not be necessary here.

I think what will happen is calling `this.closeController.abort()` will cause the `pipe` in `sink` to throw, the error will be caught and passed to `this.input.end` on line 316.

Ending the input should then cause the connection to close?  It's not clear what use a muxed connection with a closed muxer is, so that's probably good.

achingbrain (changes requested):

packages/interface-stream-muxer-compliance-tests/src/close-test.ts
```diff
@@ -67,6 +67,48 @@ export default (common: TestSetup<StreamMuxerFactory>) => {
...
+        streams.map(async stream => {
+          return await pipe(
+            infiniteRandom,
+            stream,
+            drain
+          )
+        })
+      )
+
+      expect(dialer.streams).to.have.lengthOf(expectedStreams)
```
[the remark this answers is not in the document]
wemeetagain: :stuck_out_tongue_closed_eyes:

packages/interface-mocks/src/muxer.ts
wemeetagain: Right, I think the only big difference is that by calling `this.input.end(err)` explicitly, the error is passed to the source.
If we rely on the pipe in sink to throw, the error that the source will get will be some generic "aborted" error.

Not sure which is the desired behavior, or that either behavior matters to the interface.

achingbrain: Maybe `this.input.end(err)` is better - it's not async and the received error is more useful than 'aborted'.

achingbrain (approved):

packages/interface-mocks/src/muxer.ts
```diff
@@ -399,6 +397,19 @@ class MockMuxer implements StreamMuxer {
 
     return muxedStream
   }
+
+  close (err?: Error): void {
+    if (this.closeController.signal.aborted) return
+    this.log('closing muxed streams')
+
+    if (err == null) {
```
paralin, line 405: if (err) instead to catch undefined ?

PR #255, opened 2022-06-22
feat!: add peer tagging

achingbrain:
Allow tagging peers to better prioritise which connections to kill when hitting limits.  Also for keeping "priority" connections alive.

Refs: https://github.com/libp2p/js-libp2p/issues/369

BREAKING CHANGE: implementations will need to add the various tag methods


achingbrain:
`1.1.0`? 🤨

PR #260, opened 2022-06-24
fix!: use `Set` for improved performance when searching in `multicodecs`

D4nte:
The `multicodecs` values are the protocol codecs for the given pubsub
implementation. They are likely to be set once (at instantiation) but
accessed multiple time: everytime it needs to check whether a remote
peer supports the same pubsub protocol.

Hence, the usage of `Set` over `Array` is more adapted as it will
improve access performance by using `Set.prototype.has` O(1) instead
of `Array.prototype.includes` O(N).

Refs: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288

### DIFF
--- a/packages/interface-pubsub/src/index.ts
+++ b/packages/interface-pubsub/src/index.ts
@@ -71,7 +71,7 @@ export interface PeerStreams extends EventEmitter<PeerStreamEvents> {
 export interface PubSubInit {
   enabled?: boolean
 
-  multicodecs?: string[]
+  multicodecs?: Set<string>
 
   /**
    * defines how signatures should be handled


achingbrain:
I think the `.multicodecs` property on the `PubSub` needs updating to be a `Set` as well.

I'm curious - `Set.has` is faster than `Array.includes`, but `.multicodecs` should be a very small array - have you measured a real-world difference in performance with this change?

D4nte:
> I think the [`.multicodecs` property](https://github.com/libp2p/js-libp2p-interfaces/pull/260/files#diff-25738222af29e018b858dc07d5bfb59aff8e14d48ff633658a9ca51a207d0d43R128) on the `PubSub` interface needs updating to be a `Set` as well.
> 
> I'm curious - `Set.has` is faster than `Array.includes`, but `.multicodecs` should be a very small array - have you measured a real-world difference in performance with this change?

I have not measured a real word difference at this stage.

I agree that no matter the pubsub implementation, the `multicodecs` is meant to be small. I would expect it to contain various concurrent versions of a given pubsub protocol. One is unlikely to keep too many of those around just for backward compatibility's sake.

The gain of 1 `includes` call being replaced by 1 `has`, might be marginal, but it is interesting to place it in the context of this change: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288

The protocol check is done in two loops: https://github.com/ChainSafe/js-libp2p-gossipsub/blob/475c861da7c95c4ccf6013cfa932b8fbfc9d39f3/src/index.ts#L2344

We loop on each mesh and then on each peer in the given topic (not only mesh peer, all peers in topic).

Moreover, when comparing `Set` to `Array`, there are 2 other criteria to look at apart from the search logic/performance:

- `Set` only allows unique values: This makes sense in the context of `multicodecs`.
- `Set` is less efficient than `Array` at insertion: I would expect most/all pubsub implementation to initialized `multicodecs` in the class constructor so this downside is irrelevant.

I can look into performance if we think it would provide valuable information.

achingbrain:
My feeling on this sort of stuff is that it's easy to do things in the name of performance, but this is a breaking change which will be disruptive so that comes at a cost.

If there's no real-world benefit in terms of performance, it weakens the case for making a disruptive change, so yes - I think it'd be preferable that the performance gain was measured.

No need to go to great lengths, even just seeing a difference in the time it takes to run the gossipsub test suite would be enough of an indication.

D4nte:
> No need to go to great lengths, even just seeing a difference in the time it takes to run the gossipsub test suite would be enough of an indication.

Sounds good. Will do so. At the moment, `master` of js-libp2p-gossipsub is broken due to some dep conflict. Once https://github.com/libp2p/js-libp2p-utils/pull/51 is merged and release I should be able to fix them and proceed with the benchmark.

D4nte:
I have finally done the benchmark run. tl;dr: Changing to a `Set` does bring performance improvements.

## Benchmark results

I ran js-libp2p-gossipsub benchmark suite, `npm run benchmark`

### With `multicodecs` as a `Set`

I run the benchmark locally, with the following changes applied:
- This current PR https://github.com/libp2p/js-libp2p-interfaces/pull/260
- https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288 
- https://github.com/libp2p/js-libp2p-pubsub/pull/81

Result:
```
  heartbeat
    ✔ heartbeat                                                           17.60101 ops/s    56.81491 ms/op        -         41 runs    202 s
```

### With `multicodecs` as an `Array`

I ran the benchmark including this commit: [ChainSafe/js-libp2p-gossipsub@`4b873a3` (#288)](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288/commits/4b873a3bcca1501158728d8a201ceb44f8f55017)

Result:
```
  heartbeat
    ✔ heartbeat                                                           13.40017 ops/s    74.62591 ms/op        -         14 runs    205 s
```

D4nte:
I have finally done the benchmark run.

## Benchmark results

I ran js-libp2p-gossipsub benchmark suite, `npm run benchmark` 5 times each on my laptop with no other application running (apart from system and Gnome).

### With `multicodecs` as a `Set`

I run the benchmark locally, with the following changes applied:

- This current PR https://github.com/libp2p/js-libp2p-interfaces/pull/260
- https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288 
- https://github.com/libp2p/js-libp2p-pubsub/pull/81

Results:
```
    ✔ heartbeat  17.65120 ops/s    56.65336 ms/op        -         70 runs    201 s
    ✔ heartbeat  23.03127 ops/s    43.41923 ms/op        -         78 runs    202 s
    ✔ heartbeat  28.90457 ops/s    34.59660 ms/op        -         74 runs    204 s
    ✔ heartbeat  10.84120 ops/s    92.24075 ms/op        -         74 runs    202 s
    ✔ heartbeat  18.79453 ops/s    53.20697 ms/op        -         75 runs    200 s
```

Average: 19.844554 ops/s

### With `multicodecs` as an `Array`

I ran the benchmark including this commit: [ChainSafe/js-libp2p-gossipsub@`4b873a3` (#288)](https://github.com/ChainSafe/js-libp2p-gossipsub/pull/288/commits/4b873a3bcca1501158728d8a201ceb44f8f55017)

Results:
```
    ✔ heartbeat  27.49647 ops/s    36.36831 ms/op        -         70 runs    201 s
    ✔ heartbeat  31.32797 ops/s    31.92036 ms/op        -         72 runs    201 s
    ✔ heartbeat  29.86818 ops/s    33.48045 ms/op        -         72 runs    201 s
    ✔ heartbeat  25.70503 ops/s    38.90290 ms/op        -         70 runs    202 s
    ✔ heartbeat  22.04391 ops/s    45.36400 ms/op        -         81 runs    204 s
```

Average: 27.288312 ops/s

### Conclusion

The performance seems worse with `multicodecs` as a `Set`.
However, considering how disparate the results are, I would mark this conclusion as _low confidence_.

PR #261, opened 2022-06-24
docs: update `peer-id` link to new implementation

acolytec3:
The documentation currently points to the older version of `peer-id` and should be updated to point to the currently supported version.


wemeetagain (approved):

achingbrain:
Thanks!

PR #265, opened 2022-06-28
fix: add separate type for multiplexed stream

achingbrain:
A stream only has a protocol once it's been negotiated, but that happens after the muxer has opened it, so make the protocol type required on the return type from `Connection.newStream` but not present on `StreamMuxerFactory.newStream`.

All the same types are exported so this should be non-breaking.


wemeetagain (approved):

wemeetagain:
Yeah sounds good to me

maschad:
Closing in favour of https://github.com/libp2p/js-libp2p/issues/1849

Issue #258, opened 2022-06-22
Add support for `collect`-style metrics


wemeetagain:
See https://github.com/siimon/prom-client#configuration

> If the gauge is used for a point-in-time observation, you should provide a collect function

In some cases, it makes more sense to configure metrics via collect callbacks rather than continually synchronize external gauges with internal state.

Issue #257, opened 2022-06-22
Add support for metrics labels


wemeetagain:
See https://github.com/siimon/prom-client#labels

Sometimes we want metrics with specific labels, eg:
![metrics-labels](https://user-images.githubusercontent.com/1348242/175133263-2c96c4c0-e4e8-44ed-9a96-948ae602ecf2.png)

In this example, the metric was set up and used like so:
```ts
const peerGauge = new Gauge({
  name: 'peer_count',
  help: 'peer count',
  labels: ['direction']
})

peerGauge.set({direction: 'inbound'}, 25)
peerGauge.set({direction: 'outbound'}, 25)
```

The metrics interface should support some way to do this.

achingbrain:
Sounds good.  As long as we can keep prometheus-specific stuff out of the codebase, this would be really great to have.  I'd like to graph protocol counts together as well, for example.

PR #267, opened 2022-06-30
feat!: add metrics groups

achingbrain:
To better support more sophisticated graphs, allow grouping metrics by a label and also passing a function to calculate either individual metrics or groups of metrics.

Fixes: #257
Fixes: #258

BREAKING CHANGE: the return type of `metrics.getComponentMetrics` has been changed to include optional labels/help text and also is now a function that returns a single or group value

### DIFF
--- a/packages/interface-metrics/src/index.ts
+++ b/packages/interface-metrics/src/index.ts
@@ -114,18 +114,66 @@ export interface StreamMetrics {
   trackStream: <T extends Duplex<Uint8Array>> (data: TrackStreamOptions<T>) => T
 }
 
+/**
+ * Used to update a tracked metric. Value can either be a number, an object containing
+ * key/value pairs or an (optionally async) function to return a number or an object of
+ * key/value pairs.
+ */
 export interface ComponentMetricsUpdate {
+  /**
+   * Name of the system, e.g. libp2p, ipfs, etc
+   */
   system: string
+
+  /**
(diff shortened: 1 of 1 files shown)


achingbrain:
Using this to poke around how many streams we have running:

image

wemeetagain (approved):
This looks great!

wemeetagain:
Maybe spoke too soon. Looking at how to put this into practice (eg: in js-libp2p-tracked-map) I think the `label` needs both a key and value. In your example, the key is `"protocol"`.

achingbrain:
ipfs exposes the metrics recorded by libp2p as a prometheus endpoint [like this](https://github.com/ipfs/js-ipfs/blob/feat/update-libp2p-0.38.0/packages/ipfs-http-server/src/api/routes/debug.js#L31-L69) and is how the above graph is created - maybe that's helpful?

achingbrain:
Though I guess you want to track a bunch of maps together?  In that case perhaps just use regular maps and have a [CalculateComponentMetric](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-metrics/src/index.ts#L164) that returns a [ComponentMetricsGroup](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-metrics/src/index.ts#L159) with the sizes of the various maps?

PR #275, opened 2022-08-01
feat!: change connection interface to use Uint8ArrayList

mpetrunic:
@achingbrain it seems like it-handshake is used to exchange peer id.
I hacked to not use it-handshake transform after that as it converts everything to Uint8Array.
I will try to open PR to it-handshake without breaking change


achingbrain:
Looks like a coverage file crept into this PR?

mpetrunic:
> Looks like a coverage file crept into this PR?

true, I also need to update it-handshake

mpetrunic:
closing in favour of https://github.com/libp2p/js-libp2p-interfaces/pull/278

PR #278, opened 2022-08-04
feat!: change connection encryption interface to uint8arraylist

mpetrunic:
replacing #275 

BREAKING CHANGE: connection encryption interface now accepts and yields uint8arraylist instead of uint8array

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ node_modules
 # Lock files
 package-lock.json
 yarn.lock
+.vscode
--- a/packages/interface-connection-encrypter-compliance-tests/package.json
+++ b/packages/interface-connection-encrypter-compliance-tests/package.json
@@ -143,6 +143,7 @@
     "it-pair": "^2.0.2",
     "it-pipe": "^2.0.3",
     "it-stream-types": "^1.0.4",
+    "uint8arraylist": "^2.1.1",
     "uint8arrays": "^3.0.0"
   }
 }
--- a/packages/interface-connection-encrypter-compliance-tests/src/index.ts
+++ b/packages/interface-connection-encrypter-compliance-tests/src/index.ts
@@ -9,7 +9,7 @@ import { createMaConnPair } from './utils/index.js'
 import type { TestSetup } from '@libp2p/interface-compliance-tests'
 import type { ConnectionEncrypter } from '@libp2p/interface-connection-encrypter'
 import type { PeerId } from '@libp2p/interface-peer-id'
-import type { Source } from 'it-stream-types'
+import { Uint8ArrayList } from 'uint8arraylist'
 
 export default (common: TestSetup<ConnectionEncrypter>) => {
   describe('interface-connection-encrypter compliance tests', () => {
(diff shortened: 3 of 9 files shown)


wemeetagain (approved):

achingbrain:
What's the change to`@chainsafe/libp2p-noise` that meant Multiaddrs had to start sourcing/sinking `Uint8ArrayList`s?  From what I can see it'll always emit `Uint8Array`s still.

mpetrunic:
> What's the change to`@chainsafe/libp2p-noise` that meant `MultiaddrConnection`s had to start sourcing/sinking `Uint8ArrayList`s?  From what I can see it's still `Uint8Array`s in and out because the data goes through encryption libraries which only ever accept/return `Uint8Array`s.

Its not yet updated in noise and secio, will do next week

achingbrain:
secio is long deprecated & is unsupported so there's no need to spend time on that.  The other implementation that'll need updating is [PlainText](https://github.com/libp2p/js-libp2p/blob/master/src/insecure/index.ts) but noise should be done first as it doesn't depend on libp2p.

Looking at noise, there's some [concatenation](https://github.com/ChainSafe/js-libp2p-noise/blob/master/src/encoder.ts#L36) being done during the handshake that's now unnecessary since you can write a `Uint8ArrayList` directly into [writeLP](https://github.com/ChainSafe/js-libp2p-noise/blob/3ea9f33cc198d07acb2e06fcd11aa4847f1de61d/src/handshake-ik.ts#L116), but it's probably not a bottleneck because it only happens once at the start of the connection.

[This pipeline](https://github.com/ChainSafe/js-libp2p-noise/blob/master/src/noise.ts#L253-L261) is the hot code path but from what I can see because `decryptStream` [yields values](https://github.com/ChainSafe/js-libp2p-noise/blob/3ea9f33cc198d07acb2e06fcd11aa4847f1de61d/src/crypto/streaming.ts#L33-L37) from `handshake.decrypt`, which calls [chaCha20Poly1305Decrypt](https://github.com/ChainSafe/js-libp2p-noise/blob/3ea9f33cc198d07acb2e06fcd11aa4847f1de61d/src/handshakes/abstract-handshake.ts#L66) it's only ever going to yield `Uint8Array`s so there's nothing to gain from using `Uint8ArrayList`.

What am I missing?

mpetrunic:
> [This pipeline](https://github.com/ChainSafe/js-libp2p-noise/blob/master/src/noise.ts#L253-L261) is the hot code path but from what I can see because `decryptStream` [yields values](https://github.com/ChainSafe/js-libp2p-noise/blob/3ea9f33cc198d07acb2e06fcd11aa4847f1de61d/src/crypto/streaming.ts#L33-L37) from `handshake.decrypt`, which calls [chaCha20Poly1305Decrypt](https://github.com/ChainSafe/js-libp2p-noise/blob/3ea9f33cc198d07acb2e06fcd11aa4847f1de61d/src/handshakes/abstract-handshake.ts#L66) it's only ever going to yield `Uint8Array`s so there's nothing to gain from using `Uint8ArrayList`.
> 
> What am I missing?

I think you are right. I don't think noise cares if it yields or returns Uint8Array or Uint8ArrayList. But ain't we gonna have type issues if mplex or yamux is trying to pipe Uint8ArrayList into Secured connection that accepts Uint8Arrays only?

achingbrain:
No, because everything goes through `it-length-prefixed` which [only emits `Uint8Array`s](https://github.com/alanshaw/it-length-prefixed/blob/master/src/encode.ts#L33-L45) when encoding.

mpetrunic:
> No, because everything goes through `it-length-prefixed` which [only emits `Uint8Array`s](https://github.com/alanshaw/it-length-prefixed/blob/master/src/encode.ts#L33-L45) when encoding.

ugh, should we revert this change and release then?

achingbrain:
Yes & I think the generics can be removed from `MultiaddrConnection` - e.g.

```ts
export interface MultiaddrConnection<T extends Uint8Array | Uint8ArrayList = Uint8Array> extends Duplex<T> {
```

can go back to

```ts
export interface MultiaddrConnection extends Duplex<Uint8Array> {
```

PR #290, opened 2022-09-27
Add Upgrader Options

ckousik:
Allow upgrader to accept a StreamMuxerFactory and skip encryption for transports that inherently support encryption and/or multiplexing (WebRTC, WebTransport, etc.)


ckousik:
@mxinden @marcopolo

MarcoPolo (approved):

BigLep:
Per https://github.com/libp2p/js-libp2p-interfaces/pull/292#issue-1388480558

Fixes: https://github.com/libp2p/js-libp2p/issues/1382 and https://github.com/libp2p/js-libp2p-interfaces/issues/287

Enables WebTransport: https://github.com/libp2p/js-libp2p-webtransport/

packages/interface-transport/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@libp2p/interface-transport",
-  "version": "1.0.4",
+  "version": "1.0.5",
```
MarcoPolo, line 3:
```suggestion
  "version": "1.0.4",
```

MarcoPolo (approved):

packages/interface-stream-muxer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@libp2p/interface-stream-muxer",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
MarcoPolo, line 3: Should we avoid this too for the same reason as: https://github.com/libp2p/js-libp2p-interfaces/pull/290#discussion_r987048662 ?

ckousik: ah, sorry. Will fix

ckousik:
@MarcoPolo wouldn't we have to update the patch version since we have added features?

MarcoPolo:
@achingbrain made it sound like the version numbers are handled by CI during release. I don't know much more.

packages/interface-stream-muxer/src/index.ts
```diff
@@ -18,7 +18,7 @@ export interface StreamMuxer extends Duplex<Uint8Array> {
    * Initiate a new stream with the given name. If no name is
    * provided, the id of the stream will be used.
    */
-  newStream: (name?: string) => Stream
+  newStream: (name?: string) => Stream | Promise<Stream>
```
achingbrain, line 21: Based on our learnings from #295 I think this is a breaking change.

packages/interface-transport/src/index.ts
```diff
@@ -78,16 +79,21 @@ export interface UpgraderEvents {
   'connectionEnd': CustomEvent<Connection>
 }
 
+export interface UpgraderOptions {
+  skipEncryption?: boolean
+  muxerFactory?: StreamMuxerFactory | undefined
```
achingbrain, line 84: Skipping connection protection should be made explicit.

It's true the transports we have right now that have their own encrypters and muxers also don't support connection protection but that may not always be the case.

We also then won't end up deciding whether or not to protect a connection based on the presence of a muxer factory in the options as we currently do in https://github.com/libp2p/js-libp2p/pull/1395/files#diff-c199200bb38fc06cbe1654f84bb2cfaed1cf0aa9bc9ed71f646932604cbd9252R266-R272

```suggestion
  skipEncryption?: boolean
  skipProtection?: boolean
  muxerFactory?: StreamMuxerFactory | undefined
```

achingbrain (changes requested):
LGTM, just needs the extra boolean option adding.

IMPORTANT: this is a breaking change.  Before merging the following needs to happen:

1. The title of this PR needs to change from `feat: add upgrader options` to `feat!: add upgrader options`
2. The commit message needs to have the following added to it:

```
BREAKING CHANGE: the return type of StreamMuxer.newStream can now return a promise
```

Either one of the above two changes will make `semantic-release` rev the version numbers correctly when CI runs on master after merge.  Both changes will make doubly sure it doesn't miss it.

ckousik:
@achingbrain that is an older commit. I can run build, lint, and test successfully.

packages/interface-transport/src/index.ts
```diff
@@ -78,16 +79,22 @@ export interface UpgraderEvents {
   'connectionEnd': CustomEvent<Connection>
 }
 
+export interface UpgraderOptions {
+  skipEncryption?: boolean
+  skipProtection?: boolean
+  muxerFactory?: StreamMuxerFactory | undefined
```
achingbrain, line 85: If the field is specified it should have the right type and not be undefined.

```suggestion
  muxerFactory?: StreamMuxerFactory
```

achingbrain: Sorry for the final nit - I'd merge this but the PR is set to not allow pushing extra commits.

ckousik: done

ckousik:
@achingbrain Could you merge this? I don't have permissions to merge.

achingbrain:
Yes, I was just waiting for CI to be green.

achingbrain:
It's here, still not sure why it doesn't appear on this PR: https://github.com/libp2p/js-libp2p-interfaces/actions/runs/3197326618

Issue #291, opened 2022-09-27
ConnectionEncrypter should have a extensions field for Noise Extensions


MarcoPolo:
earlyData in Noise has been deprecated in favor of a structured protobuf labeled Noise Extensions: https://github.com/libp2p/specs/pull/453.

PR #293, opened 2022-09-27
feat!: Add remoteExtensions to connection-encrypter

MarcoPolo:
fixes https://github.com/libp2p/js-libp2p-interfaces/issues/291

earlyData in Noise has been deprecated in favor of a structured protobuf labeled Noise Extensions: https://github.com/libp2p/specs/pull/453.

remoteExtensions and earlyData are both Noise specific. I'm not sure what the value is by having this interface defined here. Probably what we want is a minimum interface that callers expect, but the concrete implementation can return anything as long as it meets the minimum interface. This is a bigger change than I want to take, but probably the better solution that we should do when we have other encryption modules.

BREAKING CHANGE: remoteEarlyData is replaced by generic remoteExtensions

### DIFF
--- a/packages/interface-connection-encrypter/src/index.ts
+++ b/packages/interface-connection-encrypter/src/index.ts
@@ -5,7 +5,7 @@ import type { Duplex } from 'it-stream-types'
  * A libp2p connection encrypter module must be compliant to this interface
  * to ensure all exchanged data between two peers is encrypted.
  */
-export interface ConnectionEncrypter {
+export interface ConnectionEncrypter<Extension = any> {
   protocol: string
 
   /**
@@ -13,18 +13,19 @@ export interface ConnectionEncrypter {
    * pass it for extra verification, otherwise it will be determined during
    * the handshake.
    */
-  secureOutbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection>
+  secureOutbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection<Extension>>
 
   /**
    * Decrypt incoming data. If the remote PeerId is known,
    * pass it for extra verification, otherwise it will be determined during
    * the handshake
    */
-  secureInbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection>
+  secureInbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection<Extension>>
 }
 
-export interface SecuredConnection {
+export interface SecuredConnection<E> {
   conn: Duplex<Uint8Array>
   remoteEarlyData: Uint8Array
+  remoteExtensions: E
   remotePeer: PeerId
 }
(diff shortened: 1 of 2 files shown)


packages/interface-connection-encrypter/src/index.ts
```diff
@@ -5,22 +5,23 @@ import type { Duplex } from 'it-stream-types'
...
 
   /**
    * Decrypt incoming data.
    */
-  secureInbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection>
+  secureInbound: (localPeer: PeerId, connection: Duplex<Uint8Array>, remotePeer?: PeerId) => Promise<SecuredConnection<Extension>>
 }
 
-export interface SecuredConnection {
+export interface SecuredConnection<E> {
```
MarcoPolo, line 22: I used a generic here to avoid pulling in the protobuf definition.

packages/interface-connection-encrypter/src/index.ts
```diff
@@ -5,22 +5,23 @@ import type { Duplex } from 'it-stream-types'
...
   conn: Duplex<Uint8Array>
   remoteEarlyData: Uint8Array
```
mpetrunic, line 24: Can you remove this and mark PR as a breaking change?

packages/interface-connection-encrypter/src/index.ts
mpetrunic: I don't think you need to pull protobuf definition, you can define an interface in this file:

interface NoiseExtensions {
  webtransportCerthashes: Uint8Array[]
}

packages/interface-mocks/src/connection-encrypter.ts
```diff
@@ -63,7 +63,8 @@ export function mockConnectionEncrypter () {
           conn: true
         },
         remotePeer,
-        remoteEarlyData: new Uint8Array(0)
+        remoteEarlyData: new Uint8Array(0),
```
mpetrunic, line 66: remove

packages/interface-mocks/src/connection-encrypter.ts
```diff
@@ -104,7 +105,8 @@ export function mockConnectionEncrypter () {
           conn: true
         },
         remotePeer: peerIdFromBytes(remoteId.slice()),
-        remoteEarlyData: new Uint8Array(0)
+        remoteEarlyData: new Uint8Array(0),
```
mpetrunic, line 108: remove

mpetrunic (changes requested):

packages/interface-connection-encrypter/src/index.ts
MarcoPolo: This means you have to keep this up to date with the generated code from the protobuf. Doesn't seem ideal, no?

mpetrunic: This is an interface, so it should ensure all implementations get a type error in case they are incompatible with the version of interface.

mpetrunic: Ah, on second thought, you are right, it probably makes no sense to keep this interface noise specific.
We should make `remoteExtensions` optional and change SecuredConnection into `export interface SecuredConnection<E = unknown>`

MarcoPolo: Updated. But does it make sense to set the default generic param for SecuredConnection? I don't think we need to

mpetrunic (approved):

PR #294, opened 2022-09-29
fix: make outbound peer id optional

achingbrain:
In order to support connecting to mulitaddrs without a peer id make the remote peer id an optional argument when securing the outbound connection.


mpetrunic (approved):

achingbrain:
@marin follow-up PR to noise here - https://github.com/ChainSafe/js-libp2p-noise/pull/218

PR #295, opened 2022-09-30
feat: add acceptIncomingConnection to ConnectionManager

achingbrain:
In order to check we can accept an incoming connection before exchanging PeerIds, add a method to the ConnectionManager interface that can be called by the upgrader.

### DIFF
--- a/packages/interface-connection-manager/src/index.ts
+++ b/packages/interface-connection-manager/src/index.ts
@@ -28,10 +28,10 @@ export interface ConnectionManager extends EventEmitter<ConnectionManagerEvents>
   /**
    * Invoked after an incoming connection is opened but before PeerIds are
    * exchanged, this lets the ConnectionManager check we have sufficient
-   * resources to accept the connection, otherwise it will be closed and an
-   * error thrown
+   * resources to accept the connection in which case it will return true,
+   * otherwise it will return false.
    */
-  acceptIncomingConnection: (maConn: MultiaddrConnection) => Promise<void>
+  acceptIncomingConnection: (maConn: MultiaddrConnection) => Promise<boolean>
 }
 
 export interface Dialer {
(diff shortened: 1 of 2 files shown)


packages/interface-connection-manager/src/index.ts
```diff
@@ -24,6 +24,14 @@ export interface ConnectionManager extends EventEmitter<ConnectionManagerEvents>
...
    */
   closeConnections: (peer: PeerId) => Promise<void>
+
+  /**
+   * Invoked after an incoming connection is opened but before PeerIds are
+   * exchanged, this lets the ConnectionManager check we have sufficient
+   * resources to accept the connection, otherwise it will be closed and an
+   * error thrown
+   */
+  acceptIncomingConnection: (maConn: MultiaddrConnection) => Promise<void>
```
mpetrunic, line 34: shouldn't this return a boolean or are we expecting implementation to throw?

achingbrain: The implementation throws.  It could accept a boolean but throwing lets the error be specific about the failure, though we could just log it instead?

mpetrunic: I think logging errors and returning boolean would make more sense as you cannot really handle that error up the stack other than logging it. On other hand, it's gonna be obvious to the implementer what it needs to do.

MarcoPolo:
So does this break js-libp2p until https://github.com/libp2p/js-libp2p/pull/1398 is merged?

mpetrunic:
> So does this break js-libp2p until [libp2p/js-libp2p#1398](https://github.com/libp2p/js-libp2p/pull/1398) is merged? My trivial change is breaking CI: [libp2p/js-libp2p#1398](https://github.com/libp2p/js-libp2p/pull/1398)

https://github.com/libp2p/js-libp2p/pull/1404 is merged as temporary fix

PR #296, opened 2022-10-04
fix: add default extension type

achingbrain:
Similar to the `ConnectionEncrypter` interface, specify a default type for the `Extension` so consumers uninterested in the `remoteExtensions` field can operate as before.


MarcoPolo:
ah yes, I forgot consumers have to return this type directly and it's a bit of a pain to have them plumb `<unknown>`.

MarcoPolo (approved):

PR #298, opened 2022-10-06
fix!: add topicValidators to pubsub interface

### LINKED ISSUE #297: Broken pubsub type when using external pubsub library
- **Version**: 0.39.2

- **Platform**: N/A (all platforms)

- **Subsystem**: js-libp2p or js-libp2p-interfaces

#### Description:

As described in your example [examples/pubsub/message-filtering](https://github.com/libp2p/js-libp2p/blob/0ecc02b2a426b6dfec7b6f46d565fde41ad66954/examples/pubsub/message-filtering/README.md), when using `@chainsafe/libp2p-gossipsub`, you should be able to use `node.pubsub.topicValidators` in order to pass a function which can prevent unwanted messages from being spread. Although the `@chainsafe/libp2p-gossipsub` library does implement this feature correctly, TypeScript does not recognize this feature because your types are hardcoded instead of allowing the external library to pass their types.

This is confirmed [here](https://github.com/libp2p/js-libp2p/blob/0ecc02b2a426b6dfec7b6f46d565fde41ad66954/src/libp2p.ts#L59), where your pubsub type is assigned to the Libp2pNode class, that type definition is imported from `@libp2p/interface-pubsub` [here](https://github.com/libp2p/js-libp2p/blob/0ecc02b2a426b6dfec7b6f46d565fde41ad66954/src/libp2p.ts#L35), which is defined [here](https://github.com/libp2p/js-libp2p-interfaces/blob/399544b4e90ccc10711e86989d748b3751d52765/packages/interface-pubsub/src/index.ts#L136).

#### Steps to reproduce the error:

- Try out [your example code](https://github.com/libp2p/js-libp2p/blob/0ecc02b2a426b6dfec7b6f46d565fde41ad66954/examples/pubsub/message-filtering/1.js) for message filtering on pubsub, using TypeScript instead of JavaScript.

mpetrunic:
resolves #297 

- topicValidation is indeed specified in https://github.com/libp2p/specs/blob/master/pubsub/README.md#topic-validation

Still need to:
- [x] test against https://github.com/libp2p/js-libp2p-floodsub to check new compliance test works

### DIFF
--- a/packages/interface-pubsub-compliance-tests/src/api.ts
+++ b/packages/interface-pubsub-compliance-tests/src/api.ts
@@ -5,13 +5,12 @@ import pWaitFor from 'p-wait-for'
 import { fromString as uint8ArrayFromString } from 'uint8arrays/from-string'
 import delay from 'delay'
 import type { TestSetup } from '@libp2p/interface-compliance-tests'
-import { PubSub, TopicValidatorResult, Message } from '@libp2p/interface-pubsub'
+import type { PubSub } from '@libp2p/interface-pubsub'
 import type { PubSubArgs } from './index.js'
 import type { Components } from '@libp2p/components'
 import { createComponents } from './utils.js'
 import { isStartable, start, stop } from '@libp2p/interfaces/startable'
 import { mockNetwork } from '@libp2p/interface-mocks'
-import type { PeerId } from '@libp2p/interface-peer-id'
 
 const topic = 'foo'
 const data = uint8ArrayFromString('bar')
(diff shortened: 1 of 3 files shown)


mpetrunic:
cc @tuyen @dapplion I've dropped topic from params in topic validator function as you could do something like:
`topicValidators.set(topic, createTopicValidator(topic))`

mpetrunic:
> What's the reasoning behind this being a field on the public API and not an implementation-specific field?
> 
> Given that it's part of the public API and `Map`s can be mutated in place are we ok with validators being added and removed at runtime?

Spec says it's mandatory for every implementation so it makes sense to have it in public API and shared between implementations. From what I see, none of the implementations are caching it and support changing it at runtime (which spec doesn't say it must be supported):
- https://github.com/libp2p/js-libp2p-pubsub/blob/master/src/index.ts#L572
- https://github.com/ChainSafe/js-libp2p-gossipsub/blob/master/src/index.ts#L1147

mpetrunic:
If we aren't gonna adopt this in interfaces, we should remove/change example which says this is possible: https://github.com/libp2p/js-libp2p/blob/master/examples/pubsub/message-filtering/README.md

mpetrunic:
> No, I guess this just got missed during some of the refactorings.
> 
> If this works out of the box with floodsub and gossipsub it's probably safe to release it as a minor, right?

Both floodsub/pubsub and gossipsup has a different implementation. I took a bit of both.
floodsub - expects validator function to throw error and doesn't pass peerId as param
gossipsup - returns accept/ignore/reject, but passes in topic param unnessesary

achingbrain (approved):
LGTM, just a small nit about the comments

mpetrunic:
We should probably allow @wemeetagain @dapplion and @tuyennhv as this impacts gossipsub

achingbrain:
@mpetrunic could you PR floodsub and gossipsub with the updates necessary to support this interface change?

mpetrunic:
> @mpetrunic could you PR floodsub and gossipsub with the updates necessary to support this interface change?

already have changes locally for pubsub/floodsub, will do gossipsub

dapplion (approved):
Requiring functions to throw is a bad pattern I've been regreting in Lodestar for a while. Exceptions should be exceptional, and a message getting a REJECT, IGNORE is expected behaviour.

topicValidators have to be mutable at runtime too.

So these changes LGTM

tuyennhv (approved):

PR #314, opened 2022-11-22
feat: add CodeError

tabcat:
Adds [CodeError](https://github.com/libp2p/js-libp2p/issues/1269) to errors.ts

Related: https://github.com/libp2p/js-libp2p-crypto/pull/284#issuecomment-1324005434

Added public to props and code to match AbortCode style.

Also recently added the custom [Error.name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) property to [code-err](https://github.com/tabcat/code-err) implementation. Can remove it from this PR if not wanted.

@wemeetagain

### DIFF
--- a/packages/interfaces/src/errors.ts
+++ b/packages/interfaces/src/errors.ts
@@ -17,3 +17,18 @@ export class AbortError extends Error {
     return 'aborted'
   }
 }
+
+export class CodeError<T extends Record<string, any> = Record<string, never>> extends Error {
+  public readonly props: T
+
+  constructor (
+    message: string,
+    public readonly code: string,
+    props?: T
+  ) {
+    super(message)
+
+    this.name = 'CodeError'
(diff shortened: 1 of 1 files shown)


p-shahi:
Any update on this @tabcat @wemeetagain

tabcat:
Should just need review; especially the part I added wrt the custom Error.type = 'CodeError' which is new. I needed to ping Cayman when marking it ready for review.

packages/interfaces/src/errors.ts
```diff
@@ -17,3 +17,18 @@ export class AbortError extends Error {
...
+  public readonly props: T
+
+  constructor (
+    message: string,
+    public readonly code: string,
+    props?: T
+  ) {
+    super(message)
+
+    this.name = 'CodeError'
```
achingbrain, line 31:
```suggestion
    this.name = props?.name ?? 'CodeError'
```

PR #319, opened 2022-12-14
feat: safe dispatch event

### LINKED ISSUE #317: events dispatchEvent typing
Is there any way to type check the Event.type and .detail for dispatchEvent similar to the add/removeEventListener methods?

https://github.com/libp2p/js-libp2p-interfaces/blob/913ddb522243c110318af7314015a156b8c889c1/packages/interfaces/src/events.ts#L63-L76

mpetrunic:
resolves #317

### DIFF
--- a/packages/interfaces/src/events.ts
+++ b/packages/interfaces/src/events.ts
@@ -74,6 +74,10 @@ export class EventEmitter<EventMap extends { [s: string]: any }> extends EventTa
 
     return result
   }
+
+  safeDispatchEvent<Detail>(type: keyof EventMap, detail: CustomEventInit<Detail>): boolean {
+    return this.dispatchEvent(new CustomEvent<Detail>(type as string, detail))
+  }
 }
 
 /**
--- a/packages/interface-mocks/src/connection-manager.ts
+++ b/packages/interface-mocks/src/connection-manager.ts
@@ -1,4 +1,4 @@
-import { CustomEvent, EventEmitter } from '@libp2p/interfaces/events'
+import { EventEmitter } from '@libp2p/interfaces/events'
 import type { Startable } from '@libp2p/interfaces/startable'
 import type { Connection } from '@libp2p/interface-connection'
 import type { PeerId } from '@libp2p/interface-peer-id'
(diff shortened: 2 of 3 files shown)


mpetrunic:
@achingbrain any idea whats going on with lerna/nx?

packages/interfaces/src/events.ts
```diff
@@ -74,6 +74,10 @@ export class EventEmitter<EventMap extends { [s: string]: any }> extends EventTa
 
     return result
   }
+
+  safeDispatchEvent<Detail>(type: keyof EventMap, detail: CustomEventInit<Detail>): boolean {
```
tabcat, line 78: is `detail` the best name for this param?

mpetrunic: No idea, its already ref like that in multiple places

achingbrain: `detail` is used in other places as it's the option key from the [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) constructor

achingbrain:
> any idea whats going on with lerna/nx?

Something has shifted in the dependency graph that means nx can't calculate the dependency graph as it trips over [this dependency](https://github.com/ipfs/aegir/blob/master/package.json#L225).

That dep version prevents `@semantic-release/github` hitting request limits when publishing from monorepos with lots of packages like this one - [the PR](https://github.com/semantic-release/github/pull/487) has been open since April!

packages/interfaces/src/events.ts
tabcat: right it seems confusing since this is actually the CustomEvent options object which contains detail
https://github.com/libp2p/js-libp2p-interfaces/blob/2d7e06b882bed19c0ec208d05dd6404dc3d84de1/packages/interfaces/src/events.ts#L78-L79
https://github.com/libp2p/js-libp2p-interfaces/blob/2d7e06b882bed19c0ec208d05dd6404dc3d84de1/packages/interface-mocks/src/connection-manager.ts#L97-L99

achingbrain:
If you remove the `package-lock.json` file, nx starts to work again 🙄

achingbrain (changes requested):
The idea of the typed EventEmitter here is just to be a stopgap until TypeScript adds support for typed emitters.  It's supposed to mirror the [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) interface.

This PR adds a non-standard method, `safeDispatchEvent` which stops this being a type-safe drop-in replacement for EventTarget - is there no other way?

PR #339, opened 2023-01-25
fix: Added Registrar, Dialer and Connection Manager to Libp2p object

### LINKED ISSUE #338: Registrar and ConnectionManager missing from interface-libp2p
Noticed this error upgrading to latest libp2p + gossipsub:
```
@lodestar/beacon-node: src/network/gossip/gossipsub.ts:98:11 - error TS2345: Argument of type 'Libp2p' is not assignable to parameter of type 'GossipSubComponents'.
@lodestar/beacon-node:   Type 'Libp2p' is missing the following properties from type 'GossipSubComponents': registrar, connectionManager
```

maschad:
Closes #338

### DIFF
--- a/packages/interface-libp2p/package.json
+++ b/packages/interface-libp2p/package.json
@@ -133,6 +133,7 @@
   },
   "dependencies": {
     "@libp2p/interface-connection": "^3.0.0",
+    "@libp2p/interface-connection-manager": "^1.3.7",
     "@libp2p/interface-content-routing": "^2.0.0",
     "@libp2p/interface-dht": "^2.0.0",
     "@libp2p/interface-keychain": "^2.0.0",
--- a/packages/interface-libp2p/src/index.ts
+++ b/packages/interface-libp2p/src/index.ts
@@ -15,25 +15,34 @@
  */
 
 import type { AbortOptions } from '@libp2p/interfaces'
-import type { EventEmitter } from '@libp2p/interfaces/events'
 import type { Startable } from '@libp2p/interfaces/startable'
 import type { Multiaddr } from '@multiformats/multiaddr'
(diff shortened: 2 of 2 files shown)


achingbrain (changes requested):
This is intentional to reduce the size of the libp2p api and to encourage people to use the high level methods on the libp2p node instead of reaching into the low-level components (e.g. use `libp2p.dial` instead of `libp2p.connectionManager.openConnection` etc).

maschad:
> This is intentional to reduce the size of the libp2p api and to encourage people to use the high level methods on the libp2p node instead of reaching into the low-level components (e.g. use `libp2p.dial` instead of `libp2p.connectionManager.openConnection` etc).

I understand the intention to reduce the size of the libp2p api but given that the `ConnectionManager` and `Registrar` are core to the library it would seem that those interfaces should be extended.

The impetus for this PR arose out of [instances like this](https://github.com/ChainSafe/lodestar/pull/5029/files#diff-a1d7ce3e1471462e6d30f2a0f68a99fad330bcf09dddaa4ff7f1536184ce238cR70) where more often that not a `Libp2p` instance would require being initiated with a `ConnectionManager` and `Registrar` (in this case for Gossipsub). 

Are there multiple instances where a consumer would not want to utilize the default connection manager?  
(This relates to the broader suggestion posed [here](https://github.com/libp2p/js-libp2p/issues/1567#issuecomment-1399055924))

p-shahi:
Closing per discussion in triage meeting for reasons mentioned here: https://github.com/libp2p/js-libp2p-interfaces/pull/339#pullrequestreview-1272936814

Looking into alternative options where one can optionally pass in their own connection manager module

PR #342, opened 2023-02-04
feat: Update Connection Manager and Dialler Interfaces and Configs.

maschad:
In order to avoid creating wrapper functions to extract a map of the connections from a default `ConnectionManager` implementation (as mentioned in https://github.com/libp2p/js-libp2p/issues/1567) , we should allow that field to be made accessible. 

Now whilst the `getConnections` function could achieve this, it adds [additional runtime logic which coulda affect performance](https://github.com/ChainSafe/lodestar/pull/5097#pullrequestreview-1282144695).

### DIFF
--- a/packages/interface-connection-manager/src/index.ts
+++ b/packages/interface-connection-manager/src/index.ts
@@ -1,8 +1,75 @@
 import type { AbortOptions } from '@libp2p/interfaces'
 import type { EventEmitter } from '@libp2p/interfaces/events'
-import type { Connection, MultiaddrConnection } from '@libp2p/interface-connection'
+import type { Connection, ConnectionGater, MultiaddrConnection } from '@libp2p/interface-connection'
 import type { PeerId } from '@libp2p/interface-peer-id'
-import type { Multiaddr } from '@multiformats/multiaddr'
+import type { Multiaddr, Resolver } from '@multiformats/multiaddr'
+import type { Metrics } from '@libp2p/interface-metrics'
+import type { AddressSorter, PeerStore } from '@libp2p/interface-peer-store'
+import type { TransportManager, Upgrader } from '@libp2p/interface-transport'
+export interface ConnectionManagerComponents {
+  peerId: PeerId
+  metrics?: Metrics
+  upgrader: Upgrader
+  peerStore: PeerStore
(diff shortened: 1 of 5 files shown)


packages/interface-connection-manager/src/index.ts
```diff
@@ -94,4 +105,14 @@ export interface Dialer {
...
+
+  /**
+   * Get the current dial targets which are pending
+   */
+  getPendingDialTargets: () => Map<string, AbortOptions>
+
+  /**
+   * Returns true if the peer id is in the pending dials
+   */
+  isPendingDial: (peer: PeerId | Multiaddr) => boolean
```
achingbrain, line 117:
```suggestion
  hasPendingDial: (peer: PeerId | Multiaddr) => boolean
```

achingbrain:
In order to merge this as a feature and not a breaking change a PR needs to be opened against libp2p that adds the new methods to the default connection manager.  This change then needs to ship before this PR can be merged.

maschad:
Since https://github.com/libp2p/js-libp2p/pull/1586 has been merged this can be reviewed now @achingbrain

maschad:
In light of the work being done on https://github.com/libp2p/js-libp2p/issues/1581 I have made some more updates that will be required for https://github.com/libp2p/js-libp2p/pull/1594

packages/interface-connection-manager/src/index.ts
```diff
@@ -1,8 +1,80 @@
...
+import type { TransportManager, Upgrader } from '@libp2p/interface-transport'
+export interface ConnectionManagerComponents {
+  peerId: PeerId
+  metrics?: Metrics
+  upgrader: Upgrader
+  peerStore: PeerStore
+  dialer: Dialer
+}
+
+export interface ConnectionManagerConfig {
```
achingbrain, line 17: We should remove all config interfaces from these modules, each implementation should be free to define its own config, otherwise all implementations have to use the same config which constrains what each implementation can do.

maschad:
This is no longer necessary since the refactor of the connection manager in https://github.com/libp2p/js-libp2p/pull/1678 in tandem with the changes being added in https://github.com/libp2p/js-libp2p/pull/1563

PR #351, opened 2023-03-10
feat: support batch dialling 

maschad:
See https://github.com/libp2p/js-libp2p/issues/1616 for more details

### DIFF
--- a/packages/interface-mocks/src/connection-manager.ts
+++ b/packages/interface-mocks/src/connection-manager.ts
@@ -25,9 +25,8 @@ class MockNetwork {
   }
 
   getNode (peerId: PeerId | Multiaddr []): MockNetworkComponents {
-
     if (Array.isArray(peerId)) {
-      peerId = peerIdFromString(peerId[0].getPeerId() ?? "")
+      peerId = peerIdFromString(peerId[0].getPeerId() ?? '')
     }
 
     for (const components of this.components) {
--- a/packages/interface-mocks/src/connection-manager.ts
+++ b/packages/interface-mocks/src/connection-manager.ts
@@ -77,7 +77,7 @@ class MockConnectionManager extends EventEmitter<ConnectionManagerEvents> implem
     this.started = false
   }
 
-  getConnections (peerId?: PeerId | Multiaddr[]): Connection[] {
+  getConnections (peerId?: PeerId): Connection[] {
     if (peerId != null) {
       return this.connections
         .filter(c => c.remotePeer.toString() === peerId.toString())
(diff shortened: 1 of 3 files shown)


packages/interface-mocks/src/connection-manager.ts
```diff
@@ -72,7 +77,7 @@ class MockConnectionManager extends EventEmitter<ConnectionManagerEvents> implem
     this.started = false
   }
 
-  getConnections (peerId?: PeerId): Connection[] {
+  getConnections (peerId?: PeerId | Multiaddr[]): Connection[] {
```
achingbrain, line 80: This method is from the connection manager interface and it doesn't take `Multiaddr[]` - can this be removed?

packages/interface-mocks/src/connection-manager.ts
```diff
@@ -23,7 +24,11 @@ class MockNetwork {
     this.components.push(components)
   }
 
-  getNode (peerId: PeerId): MockNetworkComponents {
+  getNode (peerId: PeerId | Multiaddr []): MockNetworkComponents {
+    if (Array.isArray(peerId)) {
+      peerId = peerIdFromString(peerId[0].getPeerId() ?? '')
```
achingbrain, line 29: Need to guard on the array length being at least 1

packages/interface-mocks/src/connection-manager.ts
```diff
@@ -23,7 +24,13 @@ class MockNetwork {
     this.components.push(components)
   }
 
-  getNode (peerId: PeerId): MockNetworkComponents {
+  getNode (peerId: PeerId | Multiaddr []): MockNetworkComponents {
+    if (Array.isArray(peerId) && peerId.length > 0) {
+      peerId = peerIdFromString(peerId[0].getPeerId() ?? '')
+    } else {
+      peerId = peerId as PeerId
```
achingbrain, line 31: You can use the [isPeerId](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-peer-id/src/index.ts#L37-L39) function exported from `@libp2p/interface-peer-id` - it tests the properties of the object, informs the compiler of the type and is safer than casting.

packages/interface-mocks/src/connection-manager.ts
```diff
@@ -90,7 +97,13 @@ class MockConnectionManager extends EventEmitter<ConnectionManagerEvents> implem
       throw new CodeError('Dialing multiaddrs not supported', 'ERR_NOT_SUPPORTED')
     }
 
-    const existingConnections = this.getConnections(peerId)
+    let existingConnections
+
+    if (Array.isArray(peerId)) {
+      existingConnections = this.getConnections(peerIdFromString(peerId[0].getPeerId() ?? ''))
```
achingbrain, line 103: This also needs guarding on then length of the array.

achingbrain (approved):

PR #364, opened 2023-04-11
fix!: remove dialer interface from interface connection manager

achingbrain:
The connection manager is how components open and close connections, the dialer is an internal implementation detail of libp2p so remove it from the interface.

BREAKING CHANGE: the Dialer interface has been removed


maschad (approved):
Once https://github.com/libp2p/js-libp2p/pull/1678 is merged we can merge this.

PR #367, opened 2023-04-11
feat: add peer:prune event to connection manager

achingbrain:
Adds a `peer:prune` event to the connection manager which can be emitted when connections are pruned.

### DIFF
--- a/packages/interface-connection-manager/src/index.ts
+++ b/packages/interface-connection-manager/src/index.ts
@@ -6,7 +6,7 @@ import type { Multiaddr } from '@multiformats/multiaddr'
 
 export interface ConnectionManagerEvents {
   /**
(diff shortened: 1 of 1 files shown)


packages/interface-connection-manager/src/index.ts
```diff
@@ -6,7 +6,7 @@ import type { Multiaddr } from '@multiformats/multiaddr'
 
 export interface ConnectionManagerEvents {
   /**
-   * This event will be triggered anytime a new Connection is established to another peer.
+   * This event will be triggered any time a new Connection is established to another peer.
```
maschad, line 9: Apparently [anytime](https://www.dictionary.com/e/anytime-vs-any-time/) is an adverb 😉

maschad (approved):
LGTM

PR #375, opened 2023-04-17
feat: allow user defined services

achingbrain:
Adds types to libp2p to allow user defined services.

Removes all non-essential services to allow them to be defined alongside any custom user versions.

This increases the modularity of libp2p as default protocols like identify, fetch and ping will not be enabled unless specifically allowed by the user, which also decreases the attack surface area of libp2p applications.

pubsub, dht, ping, fetch and identify have been removed - re-enable these by passing them as services


achingbrain:
Blocked on https://github.com/libp2p/js-libp2p/pull/1563

wemeetagain (approved):
lgtm

PR #382, opened 2023-04-21
fix: added promise to PendingDial interface

maschad:
Given that [getDialQueue](https://github.com/libp2p/js-libp2p/blob/master/src/libp2p.ts#L348) calls the internal connection manager's [getDialQueue](https://github.com/libp2p/js-libp2p/blob/master/src/connection-manager/index.ts#L602) method, which would return the current `pendingDials`,  it's expected this would include an array of `PendingDials` containing the [promises of connections](https://github.com/libp2p/js-libp2p/blob/master/src/connection-manager/dial-queue.ts#L36) , this updates the interface to reflect that.

### DIFF
--- a/packages/interface-connection-manager/src/index.ts
+++ b/packages/interface-connection-manager/src/index.ts
@@ -12,6 +12,7 @@ export interface PendingDial {
   status: PendingDialStatus
   peerId?: PeerId
   multiaddrs: Multiaddr[]
+  promise: Promise<Connection>
 }
 
 export interface ConnectionManagerEvents {


achingbrain:
Could you add a bit of context here please? What problem does this solve?

maschad:
> Could you add a bit of context here please? What problem does this solve?

My apologies, I have updated the description

achingbrain:
Not including it in the interface was somewhat intentional to make it harder for callers to modify the internal state of the queue.

This API was meant to allow inspection of the queue only.

On this basis we might be better off excluding it from the list items?

If modification of the queue is desirable one thing that might be worth adding is a cancel method to remove a pending dial from the queue and reject all promises waiting on it?

maschad:
> If modification of the queue is desirable one thing that might be worth adding is a cancel method to remove a pending dial from the queue and reject all promises waiting on it?

I agree with this, although the connection manager is ultimately responsible for handling dials, this would give consumers the option should they not want to use the default connection manager.

PR #383, opened 2023-04-22
feat!: added method to cancel a pending dial by it's id

maschad:
Follow up from https://github.com/libp2p/js-libp2p-interfaces/pull/382#issuecomment-1518692994


maschad:
Currently blocked by https://github.com/libp2p/js-libp2p/issues/1822

PR #387, opened 2023-04-26
feat: add getAddresses function to return detailed address info

achingbrain:
Returning a plain list of multiaddrs isn't enough to know what is going on with the state of a node's addresses.

This PR adds a `getAddresses` function that returns detailed info about the addresses - are they being announced, was it an observed address, are we confident that the observed address is publicly routable, etc.


maschad (approved):
Look good, we should add this method to js-libp2p first and then this can be merged.

PR #388, opened 2023-04-26
feat: add routing symbols

achingbrain:
When libp2p is [configured with arbitrary services](https://github.com/libp2p/js-libp2p/pull/1563) some of those services may be content routers or peer routers or both.  An example of this is the `@libp2p/kad-dht` module.

In order to communicate to libp2p that they can provide content/peer routing cabapiliy, add well-known symbols that libp2p can use to get references to the routing implementations.


maschad (approved):
LGTM, please fix the linter issues

PR #402, opened 2023-05-16
feat: add abstract stream class

achingbrain:
Adds an extendable stream class that implements all the logic for closing/half closing streams in a consistent way.

Largely extracted from `@libp2p/mplex` and working with `@libp2p/webrtc`.

### DIFF
--- a/packages/interface-stream-muxer/src/stream.ts
+++ b/packages/interface-stream-muxer/src/stream.ts
@@ -1,0 +1,321 @@
+import { CodeError } from '@libp2p/interfaces/errors'
+import { logger } from '@libp2p/logger'
+import { abortableSource } from 'abortable-iterator'
+import { anySignal } from 'any-signal'
+import { type Pushable, pushable } from 'it-pushable'
+import { Uint8ArrayList } from 'uint8arraylist'
+import type { Direction, Stream, StreamStat } from '@libp2p/interface-connection'
+import type { Source } from 'it-stream-types'
+
+const log = logger('libp2p:stream')
+
+const ERR_STREAM_RESET = 'ERR_STREAM_RESET'
+const ERR_STREAM_ABORT = 'ERR_STREAM_ABORT'
+const ERR_SINK_ENDED = 'ERR_SINK_ENDED'
+const ERR_DOUBLE_SINK = 'ERR_DOUBLE_SINK'
+
+export interface AbstractStreamInit {
+  /**
+   * A unique identifier for this stream
+   */
+  id: string
+
+  /**
+   * The stream direction
+   */
+  direction: Direction
+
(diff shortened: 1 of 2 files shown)


packages/interface-stream-muxer/src/stream.ts
```diff
@@ -0,0 +1,321 @@
...
+const ERR_STREAM_RESET = 'ERR_STREAM_RESET'
+const ERR_STREAM_ABORT = 'ERR_STREAM_ABORT'
+const ERR_SINK_ENDED = 'ERR_SINK_ENDED'
+const ERR_DOUBLE_SINK = 'ERR_DOUBLE_SINK'
+
+export interface AbstractStreamInit {
+  /**
+   * A unique identifier for this stream
+   */
+  id: string
```
maschad, line 21: I think a number is more appropriate for a stream's id see https://github.com/libp2p/js-libp2p-mplex/blob/master/src/stream.ts#L23, we can use a `string` for a name property.

packages/interface-stream-muxer/src/stream.ts
```diff
@@ -0,0 +1,321 @@
...
+    this.stat.timeline.closeWrite = Date.now()
+    this.sinkEnded = true
+    log.trace('%s stream %s sink end - err: %o', this.stat.direction, this.id, err)
+
+    if (err != null && this.endErr == null) {
+      this.endErr = err
+    }
+
+    if (this.sourceEnded) {
+      this.stat.timeline.close = Date.now()
```
maschad, line 132: should this be `this.stat.timeline.closeRead` ?

packages/interface-stream-muxer/src/stream.ts
```diff
@@ -0,0 +1,321 @@
...
+      this.onSinkEnd(err)
+      return
+    } finally {
+      signal.clear()
+    }
+
+    try {
+      this.sendCloseWrite()
+    } catch (err) {
+      log.trace('%s stream %s error sending close', this.stat.direction, this.id, err)
```
maschad, line 272: Do you think we should include the `direction` in these trace logs?

maschad (commented):
Looks good, just left some comments.

packages/interface-stream-muxer/src/stream.ts
achingbrain: It's in there - `this.stat.direction`

packages/interface-stream-muxer/src/stream.ts
achingbrain: This is in the `onSinkEnd` function, so we set `.close` if both the sink and the source have been closed.

maschad (approved):

Issue #356, opened 2023-03-28
Remove Stream methods from `Connection` interface


maschad:
Following up from the discussion [here](https://github.com/libp2p/js-libp2p/pull/1612#issuecomment-1487181271) the muxers should be ultimately responsible for managing streams as opposed to the connections themselves

PR #417, opened 2023-06-14
fix: removed connection methods #356

mmsaki:
the muxers should be ultimately responsible for managing streams as opposed to the connections themselves

### DIFF
--- a/packages/interface-connection/src/index.ts
+++ b/packages/interface-connection/src/index.ts
@@ -180,11 +180,6 @@ export interface Connection {
   remotePeer: PeerId
   tags: string[]
   streams: Stream[]
-
-  newStream: (multicodecs: string | string[], options?: NewStreamOptions) => Promise<Stream>
-  addStream: (stream: Stream) => void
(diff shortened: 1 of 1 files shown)


achingbrain:
Thanks for opening this, but we're about to archive this repo as the modules here will be added to the libp2p monorepo.

Could you please open a new PR against js-libp2p?

packages/interface-connection/src/index.ts
```diff
@@ -180,11 +180,6 @@ export interface Connection {
   remotePeer: PeerId
   tags: string[]
   streams: Stream[]
-
-  newStream: (multicodecs: string | string[], options?: NewStreamOptions) => Promise<Stream>
-  addStream: (stream: Stream) => void
-  removeStream: (id: string) => void
-  close: () => Promise<void>
```
achingbrain, line 187: only the `addStream` and `removeStream` methods should be deleted from this interface.

achingbrain (changes requested):

